PR #98, opened 2025-01-02
feat: implement device mode integrations

vgupta98:
## Description

This PR implements the basic structure of device mode integrations. It creates a `IntegrationPlugin` abstract class for any integration plugin to inherit and an `IntegrationsManagementPlugin` to manage and execute all the integration plugins.

## Type of change

- [ ] Documentation update

## Implementation Details

Below are the public APIs which this PR creates:
1. `addIntegration(plugin: IntegrationPlugin)` - This API adds an integration plugin to start sending event to the destination associated with that integration.
2. `removeIntegration(plugin: IntegrationPlugin)` - This API removes an integration.
3. `onDestinationReady(plugin: IntegrationPlugin, onReady: (Any?, DestinationResult) -> Unit)` - This API registers a callback to an integration plugin which is called after an (successful or failed) initialisation attempt of the destination associated with that integration.
4. `IntegrationPlugin` - an abstract class to inherit for creating an integration plugin which sends events to a destination. Below are the APIs exposed by an `IntegrationPlugin`:
    1.1 `key: String` - This key represents the key received in SourceConfig for a particular destination. This needs to be matched precisely for proper sending of events to that destination.
    1.2 `create(destinationConfig: JsonObject, analytics: Analytics, config: Configuration): Boolean` - This method needs to be overridden to instantiate/initialise the third party destination and return `true` upon its successful instantiation/initialisation.
    1.3 `getUnderlyingInstance(): Any?` - This method is overriden to return the instance of the 3rd party destination (if it exists).
    1.4 `flush()` and `reset()` - Override these to define what happens in an integration plugin when Analytics.flush or Analytics.reset are called.
    1.5 5 basic event APIs - track, screen, group, identify, alias - Override them to define their behaviour and call the corresponding APIs of the third party destination.
    1.6 `add(plugin: Plugin)` - adds a plugin to that specific integration for processing of its events.
    1.7 `remove(plugin: Plugin)` - removes a plugin from that integration.
    1.8 `teardown()` - Override this cleanup any resources when this integration plugin is removed (using removeIntegration).

#### IntegrationsManagementPlugin:
This plugin has 3 main responsibilities:
1. To host all the integration plugins in its `integrationPluginChain`.
2. To queue the events temporarily in `queuedEventsChannel` until the SourceConfig is fetched and then process those events to all the added integration plugins.
3. To register all the callbacks (registered using `Analytics.onDestinationReady(plugin: IntegrationPlugin, onReady: (Any?, DestinationResult) -> Unit)`) in `destinationReadyCallbacks` and call them at appropriate time.

Let's go through its working in detail:
1. In its `setup` - this plugin starts listening to the `sourceConfigState` in Analytics. Once fetched, it initialises all the integration plugins (added until that point of time), call the callbacks (if registered) and starts processing the events queued in `queuedEventsChannel`.
2. The `intercept` method for this plugin just queues the events in the `queuedEventsChannel` until the max limit of the channel is reached. It then drops the oldest event and queue the new event.
3. The `integrationPluginChain` is a `PluginChain` instance which holds only the integration plugins. We add plugins to it using `addIntegration` method. If the plugin is added after the SourceConfig is fetched, then `addIntegration` also calls `initAndNotifyCallbacks` which initialises and notifies the callback for the added integration plugin.
4. The `onDestinationReady` method registers a callback for an integration plugin by adding it to `destinationReadyCallbacks` map. It also checks if the plugin is already is present in the pluginChain and if that plugin is already initialised, calls the callback with `Failed` or `Ready` state.

#### IntegrationPlugin:
This plugin is the base abstract class for all the device mode integrations. As mentioned earlier - it exposes multiple APIs to the clients for defining the behaviour of third party destinations and how to instantiate it.
1. The `create` method defines the instantiation of the destination and is called inside the `initialize` method. The `initialize` is called by the `IntegrationsManagementPlugin` once the sourceConfig is fetched. It also has `destinationConfig` as a param which contains any config params provided by the sourceConfig. Client will use that to instantiate the destination.
2. The `initialize` method does the actual initialisation of destination. First, it finds the destination's config params in sourceConfig, then upon finding, it calls the `create` method. The result of `create` method determines the state of destination. It also then applies the plugins added in `pluginList` to the `pluginChain` instance which will start processing the events.
3. The `getUnderlyingInstance` returns the instance of the destination sdk.
4. The `add` method adds a plugin to a `pluginList` temporarily until the destination moves to a ready state. When the destination move to the ready state, we add these plugins from `pluginList` to the `pluginChain`.

#### Testing
For tests, following classes are created - 
1. `MockDestinationSdk` - represents a third party mock destination.
2. `MockDestinationIntegrationPlugin` - represents an integration plugin which will send events to `MockDestinationSdk`. This plugin is used to test the `IntegrationPlugin` since it is an abstract class and cannot be instantiated directly.
3. `MockDestinationCustomPlugin` - This a plugin used for event processing before sending an event to the `MockDestinationIntegrationPlugin`. It is also used in `IntegrationPluginTest`. 

## Checklist

## How to test?

Run the Android sample app and check for the logs of `SampleIntegrationPlugin`.

## Maintainers Checklist

- [ ] CI tests have passed.


android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    override suspend fun intercept(event: Event): Event? {
+        LoggerAnalytics.debug("DeviceModeDestinationPlugin: queueing event")
+        val queuedEventResult = queuedEventsChannel.trySend(event)
+
+        if (queuedEventResult.isFailure) {
+            queuedEventsChannel.tryReceive()
+            queuedEventsChannel.trySend(event)
+        }
```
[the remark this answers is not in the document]
1abhishekpandey: After discussing it offline, I got to know that this is needed to maintain the events in the in-memory queue. We mutually decided to move this into a separate method for better readability.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,133 @@
...
+    final override lateinit var analytics: Analytics
+
+    abstract val key: String
+    var isDestinationReady: Boolean = false
+        private set
+
+    private var isDestinationDisabledInSource: Boolean = false
+
+    private lateinit var pluginChain: PluginChain
+    private val pluginList: MutableList<Plugin> = mutableListOf()
```
1abhishekpandey, line 35: Let's use `destination` terminology to signify the actual use case:
```suggestion
    private val destinationPluginList: MutableList<Plugin> = mutableListOf()
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+ * will host all the device mode destination plugins in its PluginChain instance.
+ */
+internal class DeviceModeDestinationPlugin : Plugin {
+
+    override val pluginType: Plugin.PluginType = Plugin.PluginType.Destination
+
+    override lateinit var analytics: Analytics
+
+    private lateinit var pluginChain: PluginChain
+    private val pluginList: MutableList<DestinationPlugin> = mutableListOf()
```
1abhishekpandey, line 26: Let's use `destination` terminology to signify the actual use case:

```suggestion
    private lateinit var destinationPluginChain: PluginChain
    private val destinationPluginList: MutableList<DestinationPlugin> = mutableListOf()
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+            analytics.sourceConfigState
+                .filter { it.source.isSourceEnabled }
+                .first()
+                .let { sourceConfig ->
+                    pluginList.forEach { plugin ->
+                        // only destination plugins are added here
+                        pluginChain.add(plugin)
+                        // device mode destination SDKs are initialized here
+                        plugin.initialize(sourceConfig)
+                    }
```
1abhishekpandey, line 43:
```suggestion
                    destinationPluginList.forEach { destinationPlugin ->
                        destinationPluginChain.add(destinationPlugin)
                        // device mode destination SDKs are initialized here
                        destinationPlugin.initialize(sourceConfig)
                    }
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+        return event
+    }
+
+    override fun teardown() {
+        pluginChain.removeAll()
+        queuedEventsChannel.cancel()
+        pluginList.clear()
+    }
+
+    fun add(plugin: DestinationPlugin) {
```
1abhishekpandey, line 67: Make it more explicit:

```suggestion
    fun addDestination(plugin: DestinationPlugin) {
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+        pluginList.add(plugin)
+    }
+
+    fun remove(plugin: DestinationPlugin) {
```
1abhishekpandey, line 71: Make it more explicit:

```suggestion
    fun removeDestination(plugin: DestinationPlugin) {
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
1abhishekpandey: Also, make it internal:

```suggestion
    internal fun add(plugin: DestinationPlugin) {
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
1abhishekpandey: Also, make it internal:

```suggestion
    internal fun remove(plugin: DestinationPlugin) {
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+        pluginList.remove(plugin)
+        pluginChain.remove(plugin)
+    }
+
+    fun reset() {
```
1abhishekpandey, line 76: Make it `internal`:

```suggestion
    internal fun reset() {
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+                    LoggerAnalytics.debug(
+                        "DeviceModeDestinationPlugin: Destination ${plugin.key} is " +
+                            "not ready yet. Reset discarded."
+                    )
+                }
+            }
+        }
+    }
+
+    fun flush() {
```
1abhishekpandey, line 91: Make it `internal`:

```suggestion
    internal fun flush() {
```

1abhishekpandey (commented):
As per the current aligned objective, the current integration architecture looks good to me.

I'll review it again once all the comments are addressed.

Also, I would suggest you to refer to the team (internally) to understand where we need to use the `integration` and `destination` words. Currently, I've not added any comments related to that. If team aligns with the current terminology that is being used then it's fine.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.rudderstack.sdk.kotlin.core.internals.models.IdentifyEvent
+import com.rudderstack.sdk.kotlin.core.internals.models.ScreenEvent
+import com.rudderstack.sdk.kotlin.core.internals.models.SourceConfig
+import com.rudderstack.sdk.kotlin.core.internals.models.TrackEvent
+import com.rudderstack.sdk.kotlin.core.internals.plugins.Plugin
+import com.rudderstack.sdk.kotlin.core.internals.plugins.PluginChain
+import kotlinx.serialization.json.JsonObject
+
+@Suppress("TooManyFunctions")
+abstract class DestinationPlugin : Plugin {
```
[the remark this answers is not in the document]
vgupta98: I checked the EventPlugin code. We don't really need the `EventPlugin` interface. Our `RudderStackDataplanePlugin` doesn't really need EventPlugin interface, it can inherit `Plugin` interface instead.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    final override val pluginType: Plugin.PluginType = Plugin.PluginType.Destination
+
+    final override lateinit var analytics: Analytics
+
+    abstract val key: String
+    var isDestinationReady: Boolean = false
```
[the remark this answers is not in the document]
vgupta98: The `isDestinationReady` variable can be used by client to check if a destination is ready or not. Cannot think of a specific use case for now.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
vgupta98: This is not really a list of destination plugins, but actually a list of generic plugins, hence the naming is appropriate.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,133 @@
...
+        private set
+
+    private var isDestinationDisabledInSource: Boolean = false
+
+    private lateinit var pluginChain: PluginChain
+    private val pluginList: MutableList<Plugin> = mutableListOf()
+
+    protected open fun create(destinationConfig: JsonObject, analytics: Analytics, config: Configuration): Any? {
+        return null
+    }
```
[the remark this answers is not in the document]
vgupta98: There might be some SDKs which really just provide a static instance and don't really need initialisation.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,133 @@
...
+    internal fun initialize(sourceConfig: SourceConfig) {
+        val configDestination = findDestination(sourceConfig)
+        isDestinationDisabledInSource = configDestination?.isDestinationEnabled.isFalseOrNull()
+
+        if (isDestinationDisabledInSource) {
+            LoggerAnalytics.warn("DestinationPlugin: Destination $key is disabled.")
+            return
+        }
+        configDestination?.let {
+            val destination = create(it.destinationConfig, analytics, analytics.configuration as Configuration)
```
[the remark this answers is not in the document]
vgupta98: If the `create` returns null, then the `onDestinationReady` will be called with null value. It is the responsibility of the client to return a non null value if the 3rd party destination SDK provides some instance to be used for sending payloads.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,133 @@
...
+
+        when (processedEvent) {
+            is TrackEvent -> track(processedEvent)
+            is ScreenEvent -> screen(processedEvent)
+            is GroupEvent -> group(processedEvent)
+            is IdentifyEvent -> identify(processedEvent)
+            is AliasEvent -> alias(processedEvent)
+        }
+        return event
+    }
```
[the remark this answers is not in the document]
vgupta98: As said in previous comment, we do not really need `EventPlugin`, our `RudderStackDataplanePlugin` also don't really need `EventPlugin`, it can inherit `Plugin` interface instead. We can remove `EventPlugin` instead.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    override fun teardown() {
+        pluginChain.removeAll()
+        pluginList.clear()
+    }
+
+    protected open fun onDestinationReady(destination: Any?) {}
```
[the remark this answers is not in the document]
vgupta98: If we remove the `EventPlugin` interface, then these scattered APIs are really just the part of `DestinationPlugin` only. Creation of a class just to wrap these APIs would be pointless.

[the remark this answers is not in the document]
vgupta98: `getUnderlyingInstance` and `onDestinationReady` are different APIs. Although, both of them can be used to get the instance of the 3rd party destination SDK, `onDestinationReady` is really just a callback which is called when the 3rd party destination is ready to accept events/payloads.
`getUnderlyingInstance` can be called manually and multiple times to get the instance of the 3rd party destination SDK, while `onDestinationReady` is only called once when the destination is ready.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
vgupta98: The `DeviceModeDestinationPlugin` class itself is internal, so it doesn't really matters.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
vgupta98: Same, the class is internal, so internal modifier on method is redundant.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
vgupta98: Same as above.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
vgupta98: Same as above.

app/src/main/java/com/rudderstack/sampleapp/analytics/customplugins/SampleAmplitudePlugin.kt
```diff
@@ -0,0 +1,55 @@
...
+
+class SampleAmplitudePlugin: DestinationPlugin() {
+
+    private lateinit var amplitudeSdk: SampleAmplitudeSdk
+
+    override val key: String
+        get() = "Amplitude"
+
+    override fun create(destinationConfig: JsonObject, analytics: Analytics, config: Configuration): Any? {
+        val apiKey = destinationConfig["apiKey"]?.toString()
```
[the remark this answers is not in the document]
vgupta98: This casting will fail as `destinationConfig["apiKey"]` a `JsonElement` which cannot be typcasted to string.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
1abhishekpandey: I'm sorry, I somehow forgot to delete this comment.

Clarification: I didn't mean to suggest that change in this class. Please ignore it.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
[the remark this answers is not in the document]
vgupta98: As discussed offline, we will consolidate the event APIs logic in `EventPlugin`.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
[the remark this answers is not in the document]
vgupta98: As discussed offline, we will make the `create` function abstract and hence mandatory.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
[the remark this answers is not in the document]
vgupta98: As discussed offline, we will create a `getUnderlyingInstance` API to achieve parity with v1 Android SDK and will move the `onDestinationReady` API to `Analytics` class.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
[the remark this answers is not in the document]
vgupta98: As discussed offline, we have made the `isDestinationReady` variable as internal. We do not want to expose any extra APIs to the clients to avoid any possible future breaking changes.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Synchronized
+    private fun invokeOnReady(plugin: DestinationPlugin) {
+        val callBacks = destinationReadyCallbacks[plugin.key]?.toList()
+        callBacks?.forEach { callback ->
+            when (plugin.destinationState) {
+                DestinationState.Ready -> invokeAndRemoveCallback(plugin, callback, Result.Success(Unit))
+                DestinationState.Failed -> invokeAndRemoveCallback(plugin, callback, Result.Failure(error = Unit))
+                DestinationState.Uninitialised -> Unit
```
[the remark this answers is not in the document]
vgupta98: My bad, I have changed the `destinationState` to `Failed` when the destination is disabled in dashboard, so the destination can never remain `Uninitialised` after the `initialise` method is called. Hence, we do nothing in `Uninitialised` state.
Once, the sourceConfig is fetched, destination is bound to move either to `Ready` or `Failed` state. I have converted the `DestinationState` enum to a sealed interface and added an `exception` param to the `Failed` type which will now be used to pass to the callback for better information about failure.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
1abhishekpandey: I agree.
Clarification: I suggested making the method internal to improve readability.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DeviceModeDestinationPlugin.kt
1abhishekpandey: I agree.
Clarification: I suggested making the method internal to improve readability.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/DestinationPlugin.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    open fun reset() {}
+
+    final override fun setup(analytics: Analytics) {
+        super.setup(analytics)
+
+        pluginChain = PluginChain().also { it.analytics = analytics }
+    }
+
+    internal fun initialize(sourceConfig: SourceConfig) {
```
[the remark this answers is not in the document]
vgupta98: Will optimise this behaviour if required in future PRs.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPlugin.kt
```diff
@@ -0,0 +1,173 @@
...
+                    LoggerAnalytics.debug(
+                        "IntegrationsManagementPlugin: Destination ${plugin.key} is " +
+                            "not ready yet. Flush discarded."
+                    )
+                }
+            }
+        }
+    }
+
+    private fun initAndNotifyReady(sourceConfig: SourceConfig, plugin: IntegrationPlugin) {
```
1abhishekpandey, line 124:
```suggestion
    private fun initAndNotifyCallbacks(sourceConfig: SourceConfig, plugin: IntegrationPlugin) {
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPlugin.kt
```diff
@@ -0,0 +1,173 @@
...
+                DestinationState.Ready -> notifyAndRemoveCallback(plugin, callback, Result.Success(Unit))
+                is DestinationState.Failed -> notifyAndRemoveCallback(
+                    plugin,
+                    callback,
+                    Result.Failure(error = destinationState.exception)
+                )
+                DestinationState.Uninitialised -> Unit
+            }
+        }
+    }
```
1abhishekpandey, line 143: We could improve the method name and a bit of code.

Regarding method name: It's just notifying the destination callbacks. Hence the term`OnDestinationReady` doesn't seem to explain the actual use case. I think we could improve it by renaming it to the `notifyDestinationCallbacks`.


```suggestion
    private fun notifyDestinationCallbacks(plugin: IntegrationPlugin) {
        val callBacks = destinationReadyCallbacks[plugin.key]?.toList()
        callBacks?.forEach { callback ->
            when (val destinationState = plugin.destinationState) {
                is DestinationState.Ready -> notifyAndRemoveCallback(plugin, callback, Result.Success(Unit))
                is DestinationState.Failed -> notifyAndRemoveCallback(
                    plugin,
                    callback,
                    Result.Failure(error = destinationState.exception)
                )
                is DestinationState.Uninitialised -> Unit
            }
        }
    }
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPlugin.kt
1abhishekpandey: `Ready` doesn't seem to explain the proper use case!

android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPluginTest.kt
```diff
@@ -0,0 +1,230 @@
...
+        runTest {
+            plugin.addIntegration(integrationPlugin)
+            mockAnalytics.sourceConfigState.dispatch(SourceConfig.UpdateAction(sourceConfigWithCorrectApiKey))
+            advanceUntilIdle()
+
+            verify(exactly = 1) { integrationPlugin.initialize(sourceConfigWithCorrectApiKey) }
+        }
+
+    @Test
+    fun `given an integration, when it is added using addIntegration after sourceConfig is fetched, then destination is initialised`() =
```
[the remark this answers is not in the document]
1abhishekpandey: Skip this for now.

android/src/test/resources/mockdestinationconfig/source_config_with_incorrect_api_key.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "enabled": true,
+    "workspaceId": "<WORKSPACE_ID>",
+    "destinations": [
+      {
+        "id": "<DESTINATION_ID>",
+        "name": "Mock Destination",
+        "enabled": true,
+        "config": {
+          "apiKey": "random-api-key"
```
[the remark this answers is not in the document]
vgupta98: The incorrect key here does not mean syntactically incorrect key. It just means a key which is different from the key configured in the destination dashboard.

android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPluginTest.kt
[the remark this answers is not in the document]
vgupta98: As discussed offline, the `findDestination` is more correct terminology here since in sourceConfig, we are calling it `destination` and this function returns the particular destination from sourceConfig.

android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPluginTest.kt
```diff
@@ -0,0 +1,294 @@
...
+            val callback = mockk<(Any?, DestinationResult) -> Unit>(relaxed = true)
+
+            plugin.setup(mockAnalytics)
+            plugin.addIntegration(integrationPlugin)
+            mockAnalytics.sourceConfigState.dispatch(SourceConfig.UpdateAction(sourceConfigWithIncorrectApiKey))
+            plugin.onDestinationReady(integrationPlugin, callback)
+            advanceUntilIdle()
+
+            verify(exactly = 1) { callback.invoke(null, ofType(Result.Failure::class) as DestinationResult) }
+        }
```
1abhishekpandey, line 178: I think this test case could be improved!

```suggestion
    fun `given an integration which will fail to initialise, when a callback is registered, then it is called with a failure exception after the initialisation attempt`() =
        runTest {
            // Arrange
            val callback = mockk<(Any?, DestinationResult) -> Unit>(relaxed = true)
            plugin.setup(mockAnalytics)
            plugin.addIntegration(integrationPlugin)
            
            // Act
            plugin.onDestinationReady(integrationPlugin, callback)
            mockAnalytics.sourceConfigState.dispatch(SourceConfig.UpdateAction(sourceConfigWithIncorrectApiKey))
            advanceUntilIdle()

            // Assert
            verify(exactly = 1) { callback.invoke(null, ofType(Result.Failure::class) as DestinationResult) }
        }
```

1abhishekpandey: Try to follow the assert(or setup), act and assert pattern. 

Currently, it seems the arrange and act phase is mixed up a bit. In future try to follow this pattern for each test.

android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPluginTest.kt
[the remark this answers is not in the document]
1abhishekpandey: Update: @vgupta98 has added a todo for the 4th one.

1abhishekpandey (approved):

ChryssaAliferi (approved):
🙌 🙌 🙌

PR #100, opened 2025-01-08
feat: implement connectivity observer

1abhishekpandey:
## Description

- In this PR, we've added a connection availability observer for the Android and core modules. 
- Android modules rely on the Android OS to handle the connectivity callback, whereas the core module defaults to always enabling network connectivity.

## Type of change

- [ ] Documentation update

### Connectivity Observer

- Defined `ConnectivitySubscriber` interface which allows subscribing to observe a network availability.
- Defined `ConnectivityObserver` and `BaseConnectivityObserver` interfaces which provide an interface for the connectivity observers in the respective modules.
- `DefaultConnectivityObserver` (core): It implements `ConnectivityObserver` which doesn't have a logic to check for network availability. It assumes the network is available.
- `AndroidConnectivityObserver` (android): It implements `ConnectivityObserver` and has the logic to detect if the connection is available or not in the Android module. If the connection is available it notifies the subscriber immediately else once it becomes available it notifies them. It handles both legacy version and compatible SDK (i.e., API version `N = 24` or above).
    - **NOTE**: It notifies the subscriber exactly once.
    - **NOTE**: In case there is any exception while checking for connection availability, it goes to the default behaviour of notifying the users immediately and assuming the network is available.
- Analytics: It extends the `BaseConnectivityObserver` and implements the `observer`, a lazy initialisation, to ensure only one observer is active at a time.
- Added some utils methods.
- Added unit test.

Flow chart:

network_request_flowchart

## Checklist

## How to test?

1. Add this code in the Analytics class:
```
fun testConnectivityObserver() {
    this.analyticsScope.launch {
        this@Analytics.immediateNotifyOrObserveConnectivity {
            println("RudderStack: Network is avaialble")
        }
    }
}
```
2. Call `testConnectivityObserver()` from the sample app.
    a. If connection is available then: `RudderStack: Network is avaialble` will be printed.
    b. If connection is not available while checking then that lambda block will be notified once network becomes available.
3. Try with multiple other scenarios!

## Maintainers Checklist

- [ ] CI tests have passed.

## Additional Context

Refer to this PR: https://github.com/rudderlabs/rudder-sdk-kotlin/pull/101 for final changes.

### DIFF
--- a/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/Analytics.kt
+++ b/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/Analytics.kt
@@ -3,6 +3,7 @@ package com.rudderstack.sdk.kotlin.android
 import android.app.Activity
 import androidx.navigation.NavController
 import androidx.navigation.NavController.OnDestinationChangedListener
+import com.rudderstack.sdk.kotlin.android.connectivity.AndroidConnectivityObserver
 import com.rudderstack.sdk.kotlin.android.logger.AndroidLogger
 import com.rudderstack.sdk.kotlin.android.plugins.AndroidLifecyclePlugin
 import com.rudderstack.sdk.kotlin.android.plugins.AppInfoPlugin
@@ -22,6 +23,7 @@ import com.rudderstack.sdk.kotlin.android.plugins.sessiontracking.SessionTrackin
 import com.rudderstack.sdk.kotlin.android.state.NavContext
 import com.rudderstack.sdk.kotlin.android.storage.provideAndroidStorage
 import com.rudderstack.sdk.kotlin.core.Analytics
+import com.rudderstack.sdk.kotlin.core.internals.connectivity.ConnectivityObserver
 import com.rudderstack.sdk.kotlin.core.internals.logger.LoggerAnalytics
 import com.rudderstack.sdk.kotlin.core.internals.platform.Platform
 import com.rudderstack.sdk.kotlin.core.internals.platform.PlatformType
@@ -77,6 +79,7 @@ class Analytics(
     internal val activityLifecycleManagementPlugin = ActivityLifecycleManagementPlugin()
     internal val processLifecycleManagementPlugin = ProcessLifecycleManagementPlugin()
     private val sessionTrackingPlugin = SessionTrackingPlugin()
+    override lateinit var connectivityObserver: ConnectivityObserver
 
     init {
         setup()
@@ -220,6 +223,8 @@ class Analytics(
         // adding lifecycle management plugins last so that lifecycle callbacks are invoked after all the observers in plugins are added.
         add(processLifecycleManagementPlugin)
         add(activityLifecycleManagementPlugin)
+
+        connectivityObserver = AndroidConnectivityObserver((configuration as Configuration).application, this)
     }
 
     override fun getPlatformType(): PlatformType = PlatformType.Mobile
--- a/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/Analytics.kt
+++ b/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/Analytics.kt
@@ -224,7 +224,7 @@ class Analytics(
         add(processLifecycleManagementPlugin)
         add(activityLifecycleManagementPlugin)
 
-        connectivityObserver = AndroidConnectivityObserver((configuration as Configuration).application, this)
+        connectivityObserver = AndroidConnectivityObserver((configuration as Configuration).application, this.analyticsScope)
     }
 
     override fun getPlatformType(): PlatformType = PlatformType.Mobile
--- a/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserver.kt
+++ b/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserver.kt
@@ -9,7 +9,6 @@ import android.content.Intent
 import android.content.IntentFilter
 import android.net.ConnectivityManager
 import android.net.Network
-import android.net.NetworkInfo
 import android.os.Build
 import androidx.annotation.VisibleForTesting
 import com.rudderstack.sdk.kotlin.android.utils.runBasedOnSDK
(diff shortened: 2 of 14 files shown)


android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserver.kt
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Default behavior when the network becomes available.
+     *
+     * Sets the network availability status to true and notifies subscribers.
+     *
+     * This behavior ensures that network-related operations are unblocked in case of issues
+     * while monitoring the network connection status.
+     */
+    private fun defaultBehaviour() {
+        networkAvailable = AtomicBoolean(true)
```
vgupta98, line 90: AtomicBoolean variables aren't supposed to be var. We should use .set and .get methods on it. Also, we aren't really using the atomicity here, we can consider making `networkAvailable` a Volatile Boolean instead.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/connectivity/ConnectivitySubscriber.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * Observes the network availability and notifies the subscriber immediately if the network is available.
+     * Otherwise, it waits for the network to be available and then notifies the subscriber.
+     *
+     * **NOTE**: Subscriber are notified exactly once.
+     *
+     * @param subscriber The subscriber to be notified when the network is available.
+     */
+    suspend fun immediateNotifyOrObserveConnectivity(subscriber: suspend () -> Unit)
```
vgupta98, line 19: Shouldn't networkAvailability be an observable state which can either be `true` or `false`? We can have a use case in which we need to do some action in case network becomes unavailable.

android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserverTest.kt
```diff
@@ -0,0 +1,185 @@
...
+    private val broadcastReceiverSubscribersSlot = slot<() -> Unit>()
+
+    private val sdkVersions = listOf(
+        Build.VERSION_CODES.N,          // Compatible SDK version
+        Build.VERSION_CODES.LOLLIPOP,   // Legacy SDK version
+    )
+
+    @Before
+    fun setup() {
+        MockKAnnotations.init(this, relaxed = true)
```
vgupta98, line 60: Just use `Dispatchers.setMain(testDispatcher)` and then we don't really need to pass testDispatcher to `runTest` function or do `testDispatcher.scheduler.advanceUntilIdle()` (instead just use `advanceUntilIdle()`).

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/Analytics.kt
```diff
@@ -220,6 +223,8 @@ class Analytics(
         // adding lifecycle management plugins last so that lifecycle callbacks are invoked after all the observers in plugins are added.
         add(processLifecycleManagementPlugin)
         add(activityLifecycleManagementPlugin)
+
+        connectivityObserver = AndroidConnectivityObserver((configuration as Configuration).application, this.analyticsScope)
```
vgupta98, line 227: We can consider initialisation of this before plugins are added as some plugins (if they want to) can use this connectivityObserver.

core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/Utils.kt
```diff
@@ -71,3 +72,19 @@ private fun String.cleanJsonString(): String {
 fun setupLogger(logger: Logger, level: Logger.LogLevel = Logger.LogLevel.VERBOSE) {
     LoggerAnalytics.setup(logger = logger, logLevel = level)
 }
+
+// As Mockk doesn't seems to support spying on lambda function, we need to create a class for the same.
+class Block {
```
vgupta98, line 77: Nice!

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/connectivity/ConnectivitySubscriber.kt
1abhishekpandey: Currently, the scope of this PR is to observe connection availability. When the network becomes available perform certain actions.

We can add the functionality to notify when a connection becomes unavailable if needed. Something like this:
```
suspend fun notifyOrAwaitDisconnection(subscriber: suspend () -> Unit)
```

I've updated the description of the PR to clarify the scope.

Please LMK if we can improve it further.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserver.kt
1abhishekpandey: Thanks, I'm going to utilise the `.set` method. I missed that.

Regarding this:
> we aren't really using the atomicity here, we can consider making networkAvailable a Volatile Boolean instead.

We need to make this variable thread-safe. As this variable is being utilised in the callbacks (e.g., refer [here](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/43707cd66d23ecd334ced2ac6ede866b28890091/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserver.kt#L130)) and also in [this](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/43707cd66d23ecd334ced2ac6ede866b28890091/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserver.kt#L111) suspend method. So having `atomic` type variable would be needed.

android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/connectivity/AndroidConnectivityObserverTest.kt
1abhishekpandey: We are not aligned on using the `setMain`. I think `.setMain shouldn't be used unless explicitly required.

1abhishekpandey:
Please refer to this new PR: https://github.com/rudderlabs/rudder-sdk-kotlin/pull/102.

PR #102, opened 2025-01-14
feat: implement connectivity observer

1abhishekpandey:
## Description

- In this PR, we've added functionality to observe for connectivity changes i.e., both network becomes available and unvailable.
- Android modules rely on the Android OS to handle the connectivity callback, whereas the core module defaults to always enabling network connectivity.

## Type of change

- [ ] Documentation update

### ConnectivityState

- It supports three actions: 
    - Default: It signifies the default state of the connection. It is used in the `core` module to always default to `true`.
    - Enable: It sets the connection state to `true` and anyone watching on this variable will be notified immediately after calling this action.
    - Disable: It sets the connection state to `false` and anyone watching on this variable will be notified immediately after calling this action.

### AndroidConnectivityObserverPlugin
- It has the logic to detect if the connection is available or not in the Android module and modify the connectivityState whenever whenever connection state changes. It handles both legacy version (i.e., from API version 21 to 23) and compatible SDK (i.e., API version `N = 24` or above).
    - **NOTE**: In case there is any exception while checking for connection availability, it sets the connection state to `true`.
- Added some utils methods.
- Added unit test.

## Checklist

## How to test?

1. Add this code in the Analytics class and call it:
```
private fun connectivityObserver() {
    this.analyticsScope.launch {
        this@Analytics.connectivityState
            // .filter { it }            // Un-comment this to get notified only when network becomes available.
            .collect { connected ->
                println("Rudderstack: Network connection is: $connected")
            }
    }
}
```
2. It'll print whenever network changes.
3. Toggle the network on and off.

## Maintainers Checklist

- [ ] CI tests have passed.

## Additional Context

I'll implement the TODO directly in the SourceConfig PR.


core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/utils/ExceptionHandlerUtils.kt
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * **NOTE**: It catches the generic `Exception` type. Use it with caution.
+ *
+ * @param block The block to be executed.
+ * @param onException The block to be executed when an exception occurs.
+ * @param onFinally The block to be executed after the main and exception blocks are executed. Default is an empty block.
+ */
+@InternalRudderApi
+@Suppress("TooGenericExceptionCaught")
+inline fun safelyExecute(block: () -> Unit, onException: (Exception) -> Unit, onFinally: () -> Unit = {}) {
```
ChryssaAliferi, line 31: I am having some doubts on the `safelyExecute` method. I understand that it encapsulates and forces the `try` / `catch` / `finally` block, but at the same time it hides the code and does not promote much readability. I need some time to see how this would work. What were you trying to achieve with this method?

ChryssaAliferi (approved):
Great effort!! 🚀

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/utils/ExceptionHandlerUtils.kt
1abhishekpandey: For this, I'm trying to achieve two use cases:

1. Encapsulate the boilerplate try-catch code.
2. In the future, when error collection tools are implemented in our SDK, we can utilize the catch block to track and notify any exceptions. This will ensure that code modifications are not required everywhere; only a single change will be needed.

e.g., something like this:
```kotlin
inline fun safelyExecute(block: () -> Unit, onException: (Exception) -> Unit, onFinally: () -> Unit = {}) {
    try {
        block()
    } catch (e: Exception) {
        onException(e)
        ExceptionTracker.notify("Some exception occurred: $e"); // We only need a single change to track and notify exceptions.
    } finally {
        onFinally()
    }
}
```

Please let me know, if it doesn't help in promoting readability then I think it would be better to directly use the try-catch block.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/models/connectivity/ConnectivityState.kt
```diff
@@ -0,0 +1,51 @@
...
+
+private const val DEFAULT_STATE = true
+private const val CONNECTION_AVAILABLE = true
+private const val CONNECTION_UNAVAILABLE = false
+
+/**
+ * State management for connectivity.
+ */
+@InternalRudderApi
+class ConnectivityState {
```
vgupta98, line 14: Improvement/Suggestion: IMO, we can consider making this an object instead of class. It would prevent repeated initialisations of this class.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/models/connectivity/ConnectivityState.kt
```diff
@@ -0,0 +1,51 @@
...
+     * Action to set the default state of connectivity.
+     * Default state is always `true`.
+     */
+    class SetDefaultStateAction : FlowAction<Boolean> {
+
+        override fun reduce(currentState: Boolean): Boolean {
+            return DEFAULT_STATE
+        }
+    }
+
```
vgupta98, line 31: Just a suggestion:
If we have simple `FlowAction`s like this one, we can have a more kotlinic way of doing things by using a property of `fun interface`:

`val setDefaultStateAction = FlowAction<Boolean> { DEFAULT_STATE }`

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/models/connectivity/ConnectivityState.kt
1abhishekpandey: We are not creating any instance of ConnectivityState.

Please refer [here](https://discuss.kotlinlang.org/t/how-can-i-create-a-static-class-in-kotlin-static-class-not-method-or-variable/7532#:~:text=As%20long%20as%20you%20mean%20static%20nested%20classes%2C%20in%20kotlin%20a%20nested%20class%20is%20static%20by%20default.%20To%20make%20it%20a%20non%2Dstatic%20inner%20class%20it%20needs%20to%20have%20a%20inner%20flag.) and [here](https://blog.logrocket.com/create-static-methods-classes-kotlin/#static-classes-in-kotlin:~:text=In%20Kotlin%2C%20however%2C%20a%20nested%20class%20is%20static%20by%20default.) for more details.
> In Kotlin, a nested class is static by default.

Please LMK if we could improve it any other way.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/models/connectivity/ConnectivityState.kt
1abhishekpandey: Nice! 
However, I believe that if we proceed with the suggested approach, it would deviate from the existing pattern of using `Action` and `reduce` terminology. This could further complicate maintaining parity between Android and iOS, as iOS developers might find this pattern confusing.

It would be better to stick to the current pattern for now, and if needed, make changes to all occurrences simultaneously.

Pleaase LMK if you still think that we should proceed with the suggested way.

PR #105, opened 2025-01-17
chore: fix failing test case and update pre-push hook

1abhishekpandey:
## Description

- In this PR, we fixed the failing test cases in the Android module.
- Update the pre-push hook to detect and block the push request if any unit test fails in any of the modules.

## Type of change

- [ ] Documentation update

### Failing test cases

- The test cases are failing due to the recent change we introduced in this [AnalyticsUtils.kt](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/develop/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/utils/AnalyticsUtils.kt#L40) file.
- [This](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/9a98f4d50074413f3b876c69797f3105d21305a0/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/utils/AnalyticsUtils.kt#L11) code requires the main thread.
```
private val MAIN_DISPATCHER = Dispatchers.Main.immediate

@DelicateCoroutinesApi
internal fun AndroidAnalytics.runOnMainThread(block: suspend () -> Unit) = analyticsScope.launch(MAIN_DISPATCHER) {
    block()
}
```
- NOTE: In the same file, we recently added [this](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/9a98f4d50074413f3b876c69797f3105d21305a0/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/utils/AnalyticsUtils.kt#L40) extension function.
- So, when we are running all the tests together and both `Analytics.runOnAnalyticsThread` and `Analytics.application` were executed in some test files, if any of them didn’t have `Dispatchers.setMain(testDispatcher)` added in the `setup` method. The test fails.
- We’ve two choices:
    - Ensure each test class that requires the above two methods, runs on `main` thread i.e., to add this `Dispatchers.setMain(testDispatcher)` line in the `setup` method of that corresponding test file.
    - Or separate those two extension functions in a different file.
- We have decided to proceed with the second approach.

### Pre-push 

- Previously, the `pre-push` hook was unable to block the push request even though it was able to detect the failing test cases on push attempt.
- We have updated the command and removed the code which was causing issue with pre-push hook i.e., `./gradlew :core:test --daemon` forced to check only for unit tests that are present and failing on the core module. We removed that code.
- Now, the test is run for the entire project including the sample app. If there's any failing test case, it should block that!
- Also, to add extra layer of safety we have added the command to clean the project and then run unit tests

## Checklist

## How to test?

- Intentionally, modify any unit test case, either present in the `android`, `core` or `sample app` module and ensure it fails.
- Run this command from the root of the module: `.git/hooks/pre-push`. It'll fail.

## Maintainers Checklist

- [ ] CI tests have passed.

- If test cases failed it'll not push the commit:
image


ChryssaAliferi (changes requested):
Instead of creating one more utilities file with ambiguous naming, let's just move:

```
internal val Analytics.application: Application
    get() = (this.configuration as Configuration).application
```

to `AndroidUtils` which is better suited as well.

ChryssaAliferi (approved):

vgupta98 (approved):

PR #107, opened 2025-01-21
feat: add EventFilteringPlugin

vgupta98:
## Description

In this PR, `EventFilteringPlugin` is added which is responsible for handling client side event filtering for integrations.

## Type of change

- [ ] Documentation update

## Implementation Details

A new plugin called `EventFilteringPlugin` is added. It is added in the `PluginChain` instance of `IntegrationPlugin` abstract class. It will be a default plugin for all the integrations since client side event filtering is a common feature for all the integrations.
Tests for `EventFilteringPlugin` are also added.

### Feature description
The `EventFilteringPlugin` is responsible for handling client side event filtering for destinations. It filters out or allows events based on `AllowList` or `DenyList` set on the dashboard.
1. Whitelist events: Only the events present in the whitelist will be allowed.
2. Blacklist events: All the events except the ones present in the blacklist will be allowed.

The plugin also dynamically updates its state when the SourceConfig is updated. It will have the latest allow or deny list configured by the client.

## Checklist

## How to test?

The sample destination in the sample app has the key `Amplitude`.
1. Just configure allowList and denyList in Client side evnet filtering on dashboard for your Amplitude destination.
4. Run the sample app.
5. Check for logs of those events. 
They should be passed or filtered out according to the logic of allowing or denying.

## Maintainers Checklist

- [ ] CI tests have passed.

## Additional Context

We are deviating from v1 Android SDK for one particular behaviour - 
In v1, the track events having empty `event` field were passed by default without applying any filtering logic on them. For this plugin, we are not giving any special treatment to events with empty `event` field. This is in alignment with JS v3 sdk.

### DIFF
--- a/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPlugin.kt
+++ b/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPlugin.kt
@@ -49,7 +49,7 @@ abstract class IntegrationPlugin : EventPlugin {
         findDestination(sourceConfig)?.let { configDestination ->
             if (!configDestination.isDestinationEnabled) {
                 LoggerAnalytics.warn(
-                    "DestinationPlugin: Destination $key is disabled in dashboard. " +
+                    "IntegrationPlugin: Destination $key is disabled in dashboard. " +
                         "No events will be sent to this destination."
                 )
                 return
@@ -62,20 +62,20 @@ abstract class IntegrationPlugin : EventPlugin {
                         analytics.configuration as Configuration
                     )
                 } catch (e: Exception) {
-                    LoggerAnalytics.error("DestinationPlugin: Error: ${e.message} initializing destination $key.")
+                    LoggerAnalytics.error("IntegrationPlugin: Error: ${e.message} initializing destination $key.")
                     false
                 }
             ) {
                 true -> {
                     destinationState = DestinationState.Ready
-                    LoggerAnalytics.debug("DestinationPlugin: Destination $key is ready.")
+                    LoggerAnalytics.debug("IntegrationPlugin: Destination $key is ready.")
                     applyDefaultPlugins()
                     applyCustomPlugins()
                 }
 
                 false -> {
                     destinationState = DestinationState.Failed
-                    LoggerAnalytics.debug("DestinationPlugin: Destination $key is not initialised.")
+                    LoggerAnalytics.debug("IntegrationPlugin: Destination $key failed to initialise.")
                 }
             }
         }
--- a/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPlugin.kt
+++ b/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPlugin.kt
@@ -12,6 +12,12 @@ import com.rudderstack.sdk.kotlin.core.internals.plugins.PluginChain
 import kotlinx.serialization.json.JsonObject
 import java.util.concurrent.CopyOnWriteArrayList
 
+/**
+ * Base plugin class for all integration plugins.
+ *
+ * An integration plugin is a plugin that is responsible for sending events directly
+ * to a 3rd party destination without sending it to Rudder server first.
+ */
 @Suppress("TooGenericExceptionCaught")
 abstract class IntegrationPlugin : EventPlugin {
 
@@ -19,6 +25,9 @@ abstract class IntegrationPlugin : EventPlugin {
 
     final override lateinit var analytics: Analytics
 
+    /**
+     * The key for the destination present in the source config.
+     */
     abstract val key: String
 
     @Volatile
@@ -28,14 +37,34 @@ abstract class IntegrationPlugin : EventPlugin {
     private lateinit var pluginChain: PluginChain
     private val pluginList = CopyOnWriteArrayList<Plugin>()
 
+    /**
+     * Creates the destination instance. Override this method for the initialisation of destination.
+     * This method must return true if the destination was created successfully, false otherwise.
+     *
+     * @param destinationConfig The configuration for the destination.
+     * @param analytics The analytics instance.
+     * @param config The configuration instance.
+     * @return true if the destination was created successfully, false otherwise.
+     */
     protected abstract fun create(destinationConfig: JsonObject, analytics: Analytics, config: Configuration): Boolean
 
+    /**
+     * Returns the instance of the destination which was created.
+     *
+     * @return The instance of the destination.
+     */
     open fun getUnderlyingInstance(): Any? {
         return null
     }
 
+    /**
+     * Override this method to control the behaviour of [Analytics.flush] for this destination.
+     */
     open fun flush() {}
 
+    /**
+     * Override this method to control the behaviour of [Analytics.reset] for this destination.
+     */
     open fun reset() {}
 
     final override fun setup(analytics: Analytics) {
@@ -91,6 +120,11 @@ abstract class IntegrationPlugin : EventPlugin {
         return event
     }
 
+    /**
+     * Override this method to cleanup any resources before this integration plugin is removed.
+     *
+     * **Note**: Calling of `super.teardown()` is recommended when overriding this method.
+     */
     override fun teardown() {
         if (destinationState.isReady()) {
             pluginChain.removeAll()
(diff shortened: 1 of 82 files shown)


1abhishekpandey (commented):
Great work!

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
```diff
@@ -0,0 +1,107 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonObject
+import java.util.concurrent.CopyOnWriteArrayList
+
+private const val WHITE_LIST_EVENTS = "whitelistedEvents"
+private const val BLACK_LIST_EVENTS = "blacklistedEvents"
+private const val EVENT_FILTERING_OPTION = "eventFilteringOption"
+
+internal class EventFilteringPlugin(private val key: String) : Plugin {
```
1abhishekpandey, line 23: Let's add documentation for this class.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
```diff
@@ -0,0 +1,107 @@
...
+                                filteringList.clear()
+                                filteringList.addAll(getEventFilteringList(option, destinationConfig))
+                            }
+                    }.orEmpty()
+                }
+        }
+    }
+
+    override suspend fun intercept(event: Event): Event? {
+        if (event !is TrackEvent) {
```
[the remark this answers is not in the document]
vgupta98: We are deviating away from the v1 behaviour as per alignment with js sdk. CC @saikumarrs

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
```diff
@@ -0,0 +1,107 @@
...
+        super.setup(analytics)
+        analytics.analyticsScope.launch {
+            analytics.sourceConfigState
+                .filter { it.source.isSourceEnabled }
+                .collect { sourceConfig ->
+                    val destinationConfig = findDestination(sourceConfig, key)?.destinationConfig
+
+                    filteringOption = destinationConfig?.let {
+                        destinationConfig.getString(EVENT_FILTERING_OPTION)
+                            .orEmpty()
```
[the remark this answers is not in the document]
vgupta98: The `safelyExecute` block is for catching generic exception, we are only concerned about catching `IllegalArgumentException` here.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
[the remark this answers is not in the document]
ChryssaAliferi: We will discuss if this is needed to exist in the event level itself.

1abhishekpandey (approved):

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
```diff
@@ -0,0 +1,147 @@
...
+ */
+internal class EventFilteringPlugin(private val key: String) : Plugin {
+
+    override val pluginType: Plugin.PluginType = Plugin.PluginType.PreProcess
+
+    override lateinit var analytics: Analytics
+
+    @Volatile
+    private var filteringOption = String.empty()
+    private val filteringList = CopyOnWriteArrayList<String>()
```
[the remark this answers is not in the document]
vgupta98: `CopyOnWriteArrayList` is required because we are modifying this list in a coroutine in `setup` method and then reading it in a different coroutine in `intercept` method. Both these coroutines can run on different threads and we can get ConcurrentModificationException when the SourceConfig is updated dynamically.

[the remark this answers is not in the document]
vgupta98: We can let go of CopyOnWriteArrayList if we use the single threaded dispatcher for the listening of config changes coroutine. Have added a todo for it.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        val eventName = event.event.trim()
+        return when {
+            shouldDropEvent(eventName) -> {
+                LoggerAnalytics.debug("EventFilteringPlugin: Dropped event '$eventName' for destination: $key")
+                null
+            }
+            else -> event
+        }
```
1abhishekpandey, line 57: NIT: We can improve the code.
```suggestion
        return when(shouldDropEvent(eventName)) {
            true -> {
                LoggerAnalytics.debug("EventFilteringPlugin: Dropped event '$eventName' for destination: $key")
                null
            }
            false -> event
        }
```

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
```diff
@@ -0,0 +1,128 @@
...
+    private fun updateFilteringConfiguration(destinationConfig: JsonObject?) {
+        filteringOption = destinationConfig?.getString(EVENT_FILTERING_OPTION).orEmpty()
+
+        if (filteringOption.isBlank()) {
+            LoggerAnalytics.error("EventFilteringPlugin: Missing event filtering option for destination: $key")
+            filteringList.clear()
+            return
+        }
+
+        filteringList.addAll(getEventFilteringList(filteringOption, destinationConfig))
```
1abhishekpandey, line 93: There's a big issue if the values change from whitelist to blacklist or vice-versa dynamically. We should first clear the existing list!

Action item:
1. Reproduce and fix the bug.
2. Write a unit test case for this bug.
3. Add this scenario as a part of the test report.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        return parseFilteredEvents(serializedEvents)
+    }
+
+    private fun parseFilteredEvents(serializedEvents: String): List<String> {
+        return try {
+            LenientJson.decodeFromString<List<FilteredEvent>>(serializedEvents).mapNotNull {
+                it.eventName.trim().takeIf { name -> name.isNotEmpty() }
+            }
```
1abhishekpandey, line 117: Nice!

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPlugin.kt
```diff
@@ -216,17 +217,14 @@ abstract class IntegrationPlugin : EventPlugin {
     }
 
     private fun applyDefaultPlugins() {
+        add(EventFilteringPlugin(key))
```
1abhishekpandey, line 220: NIT: I would suggest wrapping the addition of custom integration plugins (standard integration plugins may not be necessary as we are maintaining them) in a try/catch block to ensure there's no event loss even when some issue occurs during the `setup` process!

We can consider doing this in future, but I thought to highlight it now!

1abhishekpandey (changes requested):

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/EventFilteringPlugin.kt
vgupta98: Nice catch!

1abhishekpandey (approved):

PR #110, opened 2025-01-26
feat: add IntegrationOptionsPlugin

vgupta98:
## Description

In this PR, `IntegrationOptionsPlugin` is added which is responsible for filtering out events for integrations based on user set integration options for events.

## Type of change

- [ ] Documentation update

## Implementation Details

A new plugin called `IntegrationOptionsPlugin` is added. It is added in the `PluginChain` instance of `IntegrationPlugin` abstract class. It will be a default plugin for all the integrations since client side event filtering is a common feature for all the integrations.
The plugin checks if the destination is explicitly enabled or disabled AND if all destinations are enabled or disabled in the event's integrations and then passes or drops the event accordingly. Tests for `IntegrationOptionsPlugin` are also added.
### Logic for Integrations options:
It applies the below logic:
All -> true - Allow
All -> false - Block

All -> false && key = true - Allow
All -> false && key = false - Block

key = true - Allow
key = false - Block
Here, `All` key represents all the destinations and `key` represents the key for a specific destination. The value for `key` is given preference over value for `All`.

Note: The `integrations` is a JsonObject therefore, it can be set to any complex JsonObject. The logic for IntegrationOptionsPlugin is such that it only looks for key of the destination and `All` key to handle the filtering logic. It will pass the event if it cannot find any of the above keys.

## Checklist

## How to test?

The sample destination in the sample app has the key `Amplitude`.
1. Set the integration options for some events before sending them as below
```kotlin
analytics.track(
    name = "Sample event",
    options = RudderOption(
        integrations = buildJsonObject {
            put("Amplitude", false)
        }
    )
)
```
2. Run the sample app.
3. Check for logs of those events. 
They should be passed or filtered out according to the integration options set.

## Maintainers Checklist

- [ ] CI tests have passed.


android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/IntegrationOptionsPluginTest.kt
1abhishekpandey, line 1: Great work with the test cases!

Let's also add few of the edge cases:

(Please rephrase/refactor them if needed)
```
@Test
fun `given an event with empty integrations, when plugin's intercept called with it, then it returns that event`() =
runTest(testDispatcher) {
    val event = TrackEvent(event = "event-name", properties = emptyJsonObject)
    event.integrations = emptyJsonObject

    val result = integrationOptionsPlugin.intercept(event)

    assertEquals(event, result)
}

@Test
fun `given an event with integration field set to string type, when plugin's intercept called with it, then it returns that event`() =
runTest(testDispatcher) {
    val event = TrackEvent(event = "event-name", properties = emptyJsonObject)
    event.integrations = buildJsonObject {
        put("MockDestination", "some string value")
    }

    val result = integrationOptionsPlugin.intercept(event)

    assertEquals(event, result)
}

@Test
fun `given an event with integration field set to complex type, when plugin's intercept called with it, then it returns that event`() =
runTest(testDispatcher) {
    val event = TrackEvent(event = "event-name", properties = emptyJsonObject)
    event.integrations = buildJsonObject {
        put("MockDestination", buildJsonObject {
            put("key", "value")
        })
    }

    val result = integrationOptionsPlugin.intercept(event)

    assertEquals(event, result)
}
```

1abhishekpandey: I can see your current logic doesn't handle above-edge cases!

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/IntegrationOptionsPlugin.kt
1abhishekpandey, line 1: In the PR description, please add the implementation details and now since we are supporting the `Json` object for `integration` please highlight that!

Over here:
> 1. Set the integration options for some events before sending them.

Let's add a sample code snippet containing the `integration` object.

1abhishekpandey (changes requested):

ChryssaAliferi (commented):
Align with Abhishek's comments on this one, so feel free to apply changes and request again for a review. 

🚀

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/eventprocessing/IntegrationOptionsPlugin.kt
```diff
@@ -0,0 +1,67 @@
...
+    override lateinit var analytics: Analytics
+
+    override suspend fun intercept(event: Event): Event? {
+        val integrationOptions = event.integrations
+
+        integrationOptions.getBoolean(key)?.let { isExplicitlyEnabled ->
+            return when (isExplicitlyEnabled) {
+                true -> event
+                false -> {
+                    logDroppedEvent(event)
```
1abhishekpandey, line 45: Nice!

1abhishekpandey (approved):
Great work!

PR #114, opened 2025-01-31
feat: use property syntax all the properties that are persisted

1abhishekpandey:
## Description

- We have provided **getters** for all properties, including `userId`, `traits`, and `anonymousId`, that persist in storage (excluding `externalId` as in future we are going to remove its persistence). Also, we've added a **setter** to set only the `anonymousId`.

## Type of change

- [x] Documentation update

## Implementation Details

- We have added getters for all the properties that are persisted and setters only for the anonymousId, as userId and traits can be set while making the `identify` event. I've not provided the option to get the `externalId`, as we've decided to remove the persistence for this value in future PR.
- I've added them as an **extension function** in the `UserIdentity` class itself, where those variables have been defined. 
- They can make a call like this:
```
// Settter
RudderAnalyticsUtils.analytics.anonymousId = "Custom Anonymous ID"
// Getter
val anonymousId = RudderAnalyticsUtils.analytics.anonymousId
val userId = RudderAnalyticsUtils.analytics.userId
val traits = RudderAnalyticsUtils.analytics.traits
```
- I've added buttons to set and fetch these properties.
- Added unit tests.

## Checklist

## How to test?

- Use the above code snippet or buttons provided in the sample app and you can see the value being fetched and set.
- `anonymousId`:
    - After SDK init, get the anonymousId: It'll return the UUID v4.
    - Call shutdown, and then get the anonymousId: It'll return null.
    - Set the anonymousId, and then get the anonymousId: It'll return the custom anonymousId.
- `userId` and `traits`:
    - After SDK init, get the userId and traits: It'll return empty values.
    - Make and identify the event and set both the values, and then try to get them: It'll return the set values.
    - Call shutdown, and then get them: It'll return null.

image

## Breaking Changes

- We have removed the `getAnonymousId` and `setAnonymousId` APIs defined in the Analytics class. Now, the user needs to access this as a property. Now, this is no longer valid:
```kotlin
val anonymousId = analytics.getAnonymousId()
analytics.setAnonymousId("<custom_anonymousId>")
```

## Maintainers Checklist

- [ ] CI tests have passed.

Before:
image

After:
image

## Additional Context

NOTE: We've not added getters for externalId, as we've planned to remove storage support for that property.

### DIFF
--- a/core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/AnalyticsTest.kt
+++ b/core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/AnalyticsTest.kt
@@ -1,0 +1,121 @@
+package com.rudderstack.sdk.kotlin.core
+
+import com.rudderstack.sdk.kotlin.core.internals.models.emptyJsonObject
+import com.rudderstack.sdk.kotlin.core.internals.models.useridentity.anonymousId
+import com.rudderstack.sdk.kotlin.core.internals.models.useridentity.traits
+import com.rudderstack.sdk.kotlin.core.internals.models.useridentity.userId
+import com.rudderstack.sdk.kotlin.core.internals.storage.Storage
+import com.rudderstack.sdk.kotlin.core.internals.storage.provideBasicStorage
+import com.rudderstack.sdk.kotlin.core.internals.utils.MockMemoryStorage
+import com.rudderstack.sdk.kotlin.core.internals.utils.empty
+import io.mockk.every
+import io.mockk.mockkStatic
+import junit.framework.TestCase.assertNull
+import junit.framework.TestCase.assertTrue
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.put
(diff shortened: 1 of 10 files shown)


core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/AnalyticsTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        assertTrue(anonymousId?.matches(uuidRegex) == true)
+    }
+
+    @Test
+    fun `given custom anonymousId is set, when anonymousId is fetched, then it should return the custom anonymousId`() {
+        analytics.anonymousId = CUSTOM_ANONYMOUS_ID
+
+        val anonymousId = analytics.anonymousId
+
+        assertTrue(anonymousId == CUSTOM_ANONYMOUS_ID)
```
vgupta98, line 57: NIT: we can use `assertEquals` instead of assertTrue at some places.

vgupta98 (approved):

ChryssaAliferi (approved):

PR #116, opened 2025-02-06
feat: add firebase integration

vgupta98:
## Description

This PR creates a `FirebaseIntegration` which is an `IntegrationPlugin` used to send events to [firebase analytics](https://firebase.google.com/docs/analytics).
It also add a `remove` API in core Analytics class to remove the plugin from PluginChain.
It also removes the `addIntegration` and `removeIntegration` APIs, now, `add` API can be used to add both normal and integration plugin and `remove` API can be used to remove it.

## Type of change

- [ ] Documentation update

## Implementation Details

Firebase Analytics supports 4 of the Rudderstack APIs - `track`, `screen`, `identify` and `reset`. These are overriden for `IntegrationPlugin` and then the corresponding methods for `FirebaseAnalytics` have been called inside them. The business logic is same as the v1 Android [firebase integration](https://github.com/rudderlabs/rudder-integration-firebase-android).

## Checklist

## Maintainers Checklist

- [ ] CI tests have passed.


1abhishekpandey (approved):

ChryssaAliferi (approved):
Great effort! 🚀

integrations/firebase/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
```
ChryssaAliferi, line 6: Regarding AndroidManifest permissions, we could remove the one's that already exist in the parent AndroidManifest such as 

```
<uses-permission android:name="android.permission.INTERNET" />
<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```

But, why do we need the WAKE_LOCK ? 
`<uses-permission android:name="android.permission.WAKE_LOCK" />`
Let's try to investigate if this is really needed. If not, let's remove all the permissions in this file and keep it blank.

vgupta98: You are right. We don't need these permissions, have removed them.

PR #118, opened 2025-02-13
fix: enable dynamic updates for integration destination config

1abhishekpandey:
## Description

- Fixed the dynamic destination config update for integrations.
- In the previous implementation we noticed that there is no way to know if the destination config is updated. Therefore, we are forced to perform parsing operations on config to get the event-to-token mapping, each time a track event is made. This behaviour was unnecessary increasing the time complexity for making track events.

## Type of change

- [ ] Documentation update

## Implementation Details

- We have added an optional `update` method in the `IntegrationPlugin class. Optional because not all integrations require this feature.
- This method will be responsible only for updating the destination config.
- Two types of value can be updated: `emptyJsonObject` (e.g., when integration is disabled) and `destination config` (e.g., whenever config is emitted).
- API:
```
open fun update(destinationConfig: JsonObject) {}
```

## Checklist

## How to test?

Testing in production requires some extra preparation:
1. Run the app for the first time: Adjust should be configured, enabled and event-to-token mapping added.
2. Stop the app.
3. Change the event-to-token mapping.
4. Relaunch the app.
5. Make track events
6. When the latest SC is emitted, you will notice the updated event-to-token mapping is being used.

Alternatively, you may refer to the unit-test for this.

## Breaking Changes

Previously, there was a variable called `destinationConfig` in the `IntegrationPlugin` class, we have replaced it with the `update` method.

## Maintainers Checklist

- [ ] CI tests have passed.


ChryssaAliferi (approved):
Looks great to me 🚀

vgupta98 (approved):

PR #125, opened 2025-02-26
fix: add reset call in identify

vgupta98:
## Description

This PR adds a `reset` (with `clearAnonymousId` as `true`) call in the `identify` API when identify is called for an already identified user but with a different userId.

## Type of change

- [ ] Documentation update

## Checklist

## How to test?

- Launch the sample app, make an event and note the session id.
- Now, identify the user by calling identify and note the session id. It should be the same, given that the user was previously anonymous.
- Now, call identify with the same userId and check the session id. It should still be the same. The `anonymousId` should also not change.
- Now, call identify with a different userId. It should change both, the `anonymousId` and the `userId`.

## Maintainers Checklist

- [ ] CI tests have passed.

## Additional Context

This logic is in alignment with JS SDK.

### DIFF
--- a/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/Analytics.kt
+++ b/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/Analytics.kt
@@ -224,6 +224,10 @@ open class Analytics protected constructor(
     ) {
         if (!isAnalyticsActive()) return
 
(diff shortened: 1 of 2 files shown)


saikumarrs (approved):

vgupta98:
This PR is closed because this change will be covered along with other changes in `reset` API in [this](https://github.com/rudderlabs/rudder-sdk-kotlin/pull/129) PR.

PR #134, opened 2025-03-07
refactor: SessionTrackingPlugin

vgupta98:
## Description

In this PR, broadly, we refactor `SessionTrackingPlugin` to break it down into two classes - `SessionTrackingPlugin` and `SessionManager`.

## Type of change

- [ ] Documentation update

## Implementation Details

- `SessionManager` class is created. It is responsible for handling session variables, updating them in storage and exposing the methods to modify them from `Analytics` class which in turn expose them as public APIs.
- `SessionTrackingPlugin` is now only responsible for attaching session variables to the payload. It does that by maintaining an instance of `SessionManager` inside it.
- The `getSessionId` API is changed to `sessionId` in `Analytics`.
- A logic for detaching the `SessionTrackingObserver` is added in `SessionManager` when `endSession` is called or `startSession` is called for manual session (we don't want any observer for manual session).
- The observer is also detached in the teardown call for `SessionTrackingPlugin`.
- Tests are added for `SessionManager`. Other tests are updated.

## Checklist

## How to test?

Test the session logic as usual. Launch the app and check for automatic session variables attached in the payload. Also, check the session APIs and manual session.

## Maintainers Checklist

- [ ] CI tests have passed.


ChryssaAliferi (approved):
Seems that we have parity now with iOS too! 
👌

PR #137, opened 2025-03-13
refactor: update header based on batch anonymous id

vgupta98:
## Description

In this PR, we update the `EventQueue` class to update the header for the batch request based on anonymousId extracted from the batch. Now, each batch request will have the header fetched from the batch.

## Type of change

- [ ] Documentation update

## Implementation Details

In the upload queue, for each batch, the anonymousId is extracted and the batch request header is updated if it is different from the previous batch anonymousId.
To filter out the `anonymousId` from the batch, following regex is used:
private val ANONYMOUS_ID_REGEX = """"anonymousId"\s*:\s*"([^"]+)"""".toRegex()

It can filter out values from jsons as below: 
1. "{"userId": "12345", "anonymousId": "abc-123", "event": "test"}" -> anonymousId = abc-123
2. "{"userId": "12345", "event": "test", "anonymousId":"xyz-456"}" -> anonymousId = xyz-456
3. "{"anonymousId": "lmn-789"}" -> anonymousId = lmn-789

If the `anonymousId` key is absent in the batch json, then a random UUID is generated and returned for the header. 

## Checklist

## Maintainers Checklist

- [ ] CI tests have passed.


core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/queue/EventQueue.kt
```diff
@@ -29,6 +27,7 @@ import java.io.FileNotFoundException
 
 internal const val UPLOAD_SIG = "#!upload"
 private const val BATCH_ENDPOINT = "/v1/batch"
+private val ANONYMOUS_ID_REGEX = """"anonymousId"\s*:\s*"([^"]+)"""".toRegex()
```
1abhishekpandey, line 30: Add an explanation in the Implementation Detail section regarding this REGEX. Mention what values are allowed and not allowed, providing sample values for both. Also, specify in case of not-allowed values do we fall back to some default value.

1abhishekpandey (approved):

PR #150, opened 2025-04-10
feat: add exponential backoff

vgupta98:
## Description

This PR adds a new policy called `BackOffPolicy` and provides its implementation called `ExponentialBackOffPolicy`.

## Type of change

- [ ] Documentation update

## Implementation Details

The `BackOffPolicy` has below methods:
- `nextDelayInMillis` - it provides the next delay in milliseconds for a given backOff policy.
- `resetBackOff` - it reset the given instance of policy to the initial state.

The `ExponentialBackOffPolicy` implements this interface and provides the delays which are increasing an exponential fashion. 
`delay = interval * base ^ attempt`
Here, `attempt` is number of times the method `nextDelayInMillis` is called. `resetBackOff` resets this value to zero again.

## Checklist

## Maintainers Checklist

- [ ] CI tests have passed.

### DIFF
--- a/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
+++ b/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
@@ -3,13 +3,13 @@ package com.rudderstack.sdk.kotlin.core.internals.policies.backoff
 import java.security.SecureRandom
 import kotlin.math.pow
 
-private const val MIN_INTERVAL = 10L
-private const val MAX_INTERVAL = 60_000L
-private const val DEFAULT_INTERVAL = 3000L
+internal const val MIN_INTERVAL = 10L
+internal const val MAX_INTERVAL = 60_000L
+internal const val DEFAULT_INTERVAL = 3000L
 
-private const val MIN_BASE = 1.1
-private const val MAX_BASE = 5.0
-private const val DEFAULT_BASE = 2.0
+internal const val MIN_BASE = 1.1
+internal const val MAX_BASE = 5.0
+internal const val DEFAULT_BASE = 2.0
 
 /**
  * This class implements an exponential backoff strategy with jitter for handling retries.
--- a/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
+++ b/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
@@ -1,6 +1,5 @@
 package com.rudderstack.sdk.kotlin.core.internals.policies.backoff
 
-import com.rudderstack.sdk.kotlin.core.internals.utils.InternalRudderApi
 import java.security.SecureRandom
 import kotlin.math.pow
 
@@ -18,8 +17,7 @@ internal const val DEFAULT_BASE = 2.0
  * The next delay is calculated using the formula: delay = interval * base^attempt.
  * The delay is then adjusted with a random jitter to avoid synchronized retries.
  */
-@InternalRudderApi
-class ExponentialBackOffPolicy(
+internal class ExponentialBackOffPolicy(
     private var intervalInMillis: Long = DEFAULT_INTERVAL,
     private var base: Double = DEFAULT_BASE,
 ) : BackOffPolicy {
--- a/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
+++ b/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
@@ -3,11 +3,11 @@ package com.rudderstack.sdk.kotlin.core.internals.policies.backoff
 import java.security.SecureRandom
 import kotlin.math.pow
 
-internal const val MIN_INTERVAL = 10L
-internal const val MAX_INTERVAL = 60_000L
-internal const val DEFAULT_INTERVAL = 3000L
+internal const val MIN_MIN_DELAY_IN_MILLIS = 3000L
+internal const val MAX_MIN_DELAY_IN_MILLIS = 60_000L
+internal const val DEFAULT_INTERVAL_IN_MILLIS = 3000L
 
-internal const val MIN_BASE = 1.1
+internal const val MIN_BASE = 2.0
 internal const val MAX_BASE = 5.0
 internal const val DEFAULT_BASE = 2.0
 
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * This class implements an exponential backoff strategy with jitter for handling retries.
+ * It allows for configurable interval and base for the exponential calculation.
+ * The next delay is calculated using the formula: delay = interval * base^attempt.
+ * The delay is then adjusted with a random jitter to avoid synchronized retries.
+ */
+@InternalRudderApi
+class ExponentialBackOffPolicy(
+    private var intervalInMillis: Long = DEFAULT_INTERVAL,
```
1abhishekpandey, line 23: Rename `intervalInMillis` to `minDelayInSec` or `initialDelayInSecs`. Refer to our internal doc for reference.

`intervalInMillis` doesn't represent the exact task of this variable. 

This variable is needed to ensure that the first delay happens after a fixed interval. To better reflect the use case, we should rename it to something that helps us understand this better.

```suggestion
    private var minDelayInSec: Int = DEFAULT_DELAY_IN_SEC,
```

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
```diff
@@ -0,0 +1,55 @@
...
+import java.security.SecureRandom
+import kotlin.math.pow
+
+internal const val MIN_INTERVAL = 10L
+internal const val MAX_INTERVAL = 60_000L
+internal const val DEFAULT_INTERVAL = 3000L
+
+internal const val MIN_BASE = 1.1
+internal const val MAX_BASE = 5.0
+internal const val DEFAULT_BASE = 2.0
```
1abhishekpandey, line 13: Do the following changes:

1. Convert all the value to seconds and avoid decimal values. We can achieve the same behaviour by having the values as second and this will make our code more readable, maintainable and simpler.
2. Let's set MIN_INTERVAL to 3 seconds; that is the default value.
3. Rename the `..._INTERVAL` to `..._DELAY_IN_SEC` or `..._INITIAL_DELAY_IN_SEC`.
4. Set the `MIN_BASE` to `2`.
5. Rename the constants to signify if they are in seconds or milliseconds. The variable name should give a clear idea of the unit.

```suggestion
internal const val MIN_DELAY_IN_SEC = 3
internal const val MAX_DELAY_IN_SEC = 6
internal const val DEFAULT_DELAY_IN_SEC = 3

internal const val MIN_BASE_IN_SEC = 2
internal const val MAX_BASE_IN_SEC = 5
internal const val DEFAULT_BASE_IN_SEC = 2
```

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * This class implements an exponential backoff strategy with jitter for handling retries.
+ * It allows for configurable interval and base for the exponential calculation.
+ * The next delay is calculated using the formula: delay = interval * base^attempt.
+ * The delay is then adjusted with a random jitter to avoid synchronized retries.
+ */
+@InternalRudderApi
+class ExponentialBackOffPolicy(
+    private var intervalInMillis: Long = DEFAULT_INTERVAL,
+    private var base: Double = DEFAULT_BASE,
```
1abhishekpandey, line 24:
```suggestion
    private var baseInSec: Int = DEFAULT_BASE_IN_SEC,
```

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
```diff
@@ -0,0 +1,55 @@
...
+            base in MIN_BASE..MAX_BASE -> base
+            else -> DEFAULT_BASE
+        }
+    }
+
+    override fun nextDelayInMillis(): Long {
+        val delayInMillis = (intervalInMillis * base.pow(attempt++)).toLong()
+        val delayWithJitterInMillis = withJitter(delayInMillis)
+
+        return delayWithJitterInMillis
```
1abhishekpandey, line 44: First add this:
```
// Define a top level constant:
private const val MILLISECONDS_IN_SECOND = 1000

// Define a top level function
private fun Int.toMilliSeconds(): Long = this * MILLISECONDS_PER_SECOND
```

```suggestion
        val delayInSec = minDelayInSec * base.pow(attempt++)
        val delayWithJitterInSec: Int = withJitter(delayInSec) // This can be made more readable with the use of infix function. But I'll skip it for now.

        return delayWithJitterInSec.toMilliseconds()
```

1abhishekpandey: I've not tested the proposed solution. Do the needful to make it work.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    private fun withJitter(delayInMillis: Long): Long {
+        val jitter = random.nextInt(delayInMillis.toInt())
+        return delayInMillis + jitter
```
1abhishekpandey, line 49:
```suggestion
    private fun withJitter(delayInSec: Int): Int {
        val jitter = random.nextInt(delayInSec)
        return delayInSec + jitter
```

1abhishekpandey: We don’t need to convert the value using `.toInt()`. Keeping the values in seconds simplifies the calculation significantly.

vgupta98:
One thought: I want to keep the calculations in milliseconds as it provides more randomness (hence, less chance of overloading server due to colliding delays from multiple clients) when adding jitter and also ultimately we need the value in milliseconds (Long) when we pass it to the `delay` function in a coroutine.
For example,
Say, calculated `delay` is 4 sec, now, by adding jitter, it can be either 4, 5, 6, 7, 8. But, if we calculate it in milliseconds, then it can range from 4000 ms to 8000 ms (all the values in between).

We can use `random.nextLong()` function in the `withJitter` method to avoid converting it to `Int`.
CC @1abhishekpandey

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
vgupta98: Base does not have a unit (say, seconds or milliSeconds). And, it has to be double value if we want to use the `pow` function from kotlin, otherwise, we will have to typecast it to `Int`.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/ExponentialBackOffPolicy.kt
vgupta98: Everthing else except `BASE` is fine. Base is a unitless value.

ChryssaAliferi (approved):
👍 Looks looks great!

1abhishekpandey (approved):

PR #156, opened 2025-04-16
feat: store events in writeKey specific directories and provide an API to delete the storage

1abhishekpandey:
## Description

- In this PR, we have added the mechanism to store events in writeKey-specific directories.
- We have also added a `delete` API to delete the storage and preference file. (This will be utilised in other PR.)
    - **NOTE: This API needs to be called carefully to avoid any crash.**

Advantages of the new approach:
- Events are now grouped by `writeKey`, making it easier to identify all batches associated with a specific key.
- This organisation allows us to delete all related batch files in a single command, improving manageability and cleanup efficiency.
- As the file name length is reduced, this will improve the performance of the SDK by a very, very small margin.

## Type of change

- [ ] Documentation update

### Change in the storage Interface

- Added the following methods in `KeyValueStorage.kt` and `Storage.kt`:
```
fun delete()
```

### AndroidStorage

- Now, events will be stored under this directory: `data/user/0/<namespace>/app_rudder-android-store_<writeKey>/<FileNo>`. Refer to the screenshot (a dummy writeKey is being used to show this example)
image

- All the events will now be grouped by the writeKey.
- **Introduced a new method `delete`: This will be responsible for deleting the `SharedPreferences` file and `Event directory` completely. It is strongly recommended that this be called as part of the `shutdown` process to avoid any crash.**

### SharedPrefsStore (Android)

- Added a `delete` method to delete the SharedPreference file. 
- Two different mechanisms are being used to delete the file based on the Android version.

### BasicStorage.kt (Core)

- Now, events will be stored under this directory: `/tmp/rudderstack-analytics-kotlin_<writeKey>/events/<FileNo>`. Refer to the screenshot (a dummy writeKey is being used to show this example)
image

- All the events will now be grouped by the writeKey.
- **Introduced a new method `delete`: This will be responsible for deleting the `Preferences` file and `Event directory` completely. It is strongly recommended that this be called as part of the `shutdown` process to avoid any crash.**

### PropertiesFile.kt (Core)

- Added a `delete` method to delete the Preference file.

### EventBatchFileManager.kt (Common)

- Refactored the logic to remove the use of the `writeKey` from the event file name prefix, i.e., now the event file name will be like this: `0,1,2,3....`.

### CustomAnnotation

- Added a new annotation called `UseWithCaution`. This can be used for the delicate API to warn the developer of the SDK to use the API carefully. Currently, it's being used in the Storage class only.
- NOTE: I've purposefully skipped adding this in the [build.gradle file](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/05de5a197a839b028d993931c9c6e3138f28e252/android/build.gradle.kts#L25-L27), so that the developer would be forced to address the warning.

### CheckBuildVersionUseCase class

- I've made it a singleton class, so that it could be called without instantiating this class.
- Added a new method `isAndroidVersionNAndAbove()`.
- Refactored the existing usage of this class: `DeeplinkPlugin.kt` and `DeeplinkPluginTest.kt`.

### Unit Test

- Update the `MockMemoryStorage.kt`.
- Added test for `SharedPrefsStoreTest.kt`.
- Fixed the failing test.

## Checklist

## How to test?

1. Add this in the Analytics class `storage.delete()` and call it at some point in time. It'll delete the storage and shared preference file completely.
    a. Check the device explorer to see if the files were deleted or not.
2. After calling the above API, call shutdown and then again call the `initialise`, a new directory will be created and the event will flow normally.
    a. Check the device explorer to see on re-initialisation, a new directory is being created or not.
3. Test the changes on at least two different Android version i.e., any latest Android API and API 21 to 23.
4. Test the changes in the Core module.

## Breaking Changes

- Now the events will be stored in a new directory which has the writeKey as its suffix.

## Maintainers Checklist

- [ ] CI tests have passed.

## Additional Context

- We will utilise the `delete` API to clear the storage. This will be called when the wrong writeKey is used. 
- Note: I'm currently planning to use this in `shutdown`, somehow, but this might change. I'll evaluate this while implementing it.

### DIFF
--- a/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/DeeplinkPlugin.kt
+++ b/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/DeeplinkPlugin.kt
@@ -21,9 +21,7 @@ internal const val REFERRING_APPLICATION_KEY = "referring_application"
 internal const val URL_KEY = "url"
 internal const val DEEPLINK_OPENED_KEY = "Deep Link Opened"
 
-internal class DeeplinkPlugin(
-    private val checkBuildVersionUseCase: CheckBuildVersionUseCase = CheckBuildVersionUseCase()
-) : Plugin, ActivityLifecycleObserver {
+internal class DeeplinkPlugin : Plugin, ActivityLifecycleObserver {
 
     override val pluginType: Plugin.PluginType = Plugin.PluginType.Utility
 
@@ -66,7 +64,7 @@ internal class DeeplinkPlugin(
     }
 
     private fun Activity.getReferrerString(): String? {
-        return if (checkBuildVersionUseCase.isAndroidVersionLollipopAndAbove()) {
+        return if (CheckBuildVersionUseCase.isAndroidVersionLollipopAndAbove()) {
             this.referrer?.toString()
         } else {
             getReferrerCompatible(this)?.toString()
--- a/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/storage/AndroidStorage.kt
+++ b/android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/storage/AndroidStorage.kt
@@ -10,7 +10,7 @@ import com.rudderstack.sdk.kotlin.core.internals.storage.MAX_PAYLOAD_SIZE
 import com.rudderstack.sdk.kotlin.core.internals.storage.Storage
 import com.rudderstack.sdk.kotlin.core.internals.storage.StorageKeys
 import com.rudderstack.sdk.kotlin.core.internals.utils.UseWithCaution
-import com.rudderstack.sdk.kotlin.core.internals.utils.appendWriteKeyToDirectoryName
+import com.rudderstack.sdk.kotlin.core.internals.utils.appendWriteKey
 import com.rudderstack.sdk.kotlin.core.internals.utils.toAndroidPrefsKey
 import java.io.File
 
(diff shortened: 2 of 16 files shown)


vgupta98 (approved):
Good job!

1abhishekpandey:
> I believe we are calling the `delete` API of storage after the `shutdown` and not during `shutdown`. Let's rephrase the docs and description of this PR to state that. `shutdown` and `delete` are unrelated, they just happen to be called in succession for a particular response of `batch` and `sourceConfig` endpoints. It could be something like: **Note**: It is recommended to use this API after shutdown to ensure storage is not removed abruptly, which could lead to unexpected errors.

Can you please point out exactly where we call the `delete` API in this PR?

1abhishekpandey:
We've utilised this `delete` API in this PR: https://github.com/rudderlabs/rudder-sdk-kotlin/pull/157. Refer [here](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/4f86df7b6b97ae7e868fbdb21b5d612cb8ba5f2c/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/Analytics.kt#L323) and it's indeed part of the shutdown. Please LMK if I need to make any correction further.

vgupta98:
> We've utilised this `delete` API in this PR: #157. Refer [here](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/4f86df7b6b97ae7e868fbdb21b5d612cb8ba5f2c/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/Analytics.kt#L323) and it's indeed part of the shutdown. Please LMK if I need to make any correction further.

So, the delete API will be a part of shutdown itself as separating it will be more complicated and it is anyways called only in case of shutdown. So, let's go ahead with the current approach.

ChryssaAliferi (approved):

PR #158, opened 2025-04-17
feat: implement retry policy for retryable event upload errors

1abhishekpandey:
## Description

- In this PR, we added the ExponentialBackOff policy for retryable errors.

## Type of change

- [ ] Documentation update

### EventUploadResult.kt

- Created a sealed interface to classify NetworkErrors into three categories: Success, RetryAbleError and Non-RetryAble error.
- This is required to implement readable exponential backoff logic.
- There is a method called `NetworkResult.toEventUploadResult()` which converts the NetworkResult into the EventUploadResult.

### MaxAttemptsExponentialBackoff

- This acts as a wrapper for the `ExponentialBackoff` class.
- This class encapsulates the logic to perform maximum retry attempts and applies exponential delay and a cool-off period.
- It also introduces a reset method that needs to be called when we want to reset the max retry attempt and exponential backoff logic.

### EventUpload

- Refactored the way we prepare the initial batch:
    - Moved the logic to update the `sentAt` timestamp. This is to ensure the `sentAt` is updated whenever a retryable error occurs.
    - Removed `prepareBatch()` method: I simplified the code further by removing this method, which previously fetched the batch payload from the storage and updated the timestamp. Now, we read the file directly and `sentAt` is updated later on. Also, we still have the `catch` block to catch any error while reading the file. We delete the file if there's any exception.
- Added a `CancellationException`: This is to ensure we catch this kind of exception explicitly. This could happen when the coroutine is suspended while retrying exponential backoff, and the shutdown operation is triggered. **NOTE: We don't remove the batch from the storage in such a case. And the SDK will try to send the same batch again when the SDK is initialised next time.**
- Refactored `uploadEvents()`:
    - Now it adds the `sentAt` to the batch before uploading it to the server/dataPlaneUrl.
    - Now it handles three types of result: `Success`, `RetryAbleError` and `NonRetryAbleError`. (Previously, it didn't have the provision to handle the `RetryAbleError`.)
    - (LOGIC IS SAME) For `Success` and `NonRetryAbleError`: SDK makes only a single attempt to upload the event and perform the action based on the status code returned from the network, e.g., if it's `Success` it'll delete the event and proceed with the next batch and if it's `NonRetryAbleError` like `ERROR_400` it'll delete the event and proceed with the next batch.
    - (NEW) But if it's `RetryAbleError` (e.g., status code `5XX` or NetworkUnAvailable etc.) - It'll implement exponential backoff logic to add delay between each retry. The maximum retry attempt is 5, after which a `CoolOff` period of 30 minutes is applied and then the loop again starts from the beginning. **NOTE: We don't remove events in case of retryable error and the loop will keep on executing infinitely until it results in a `Success`.**
- Refactored `handleNonRetryAbleError()`:
    - Now it uses `NonRetryAbleEventUploadError` instead of `NetworkErrorStatus`.

### JsonSentAtUpdater

- Changed it from `class` to `object` type, as it's basically an `util` method, we don't really need to instantiate the class in order to use the method. There is no state to be maintained.
- Added a log statement for debuggability.

### Unit Tests

- Added `TestClock` method to get the current time in tests.
- Added `EventUploadResultTest`.
- Added `MaxAttemptsExponentialBackoffTest`.
- Added `EventUploadTest`.

## Checklist

## How to test?

1. Either mock the server to return the retryable error or simply turn off the internet for the simulator, it'll create retryable error.
2. Make events.
3. You can see the batch being retried with the updated `sentAt` timestamp. 
4. Now, ensure the server returns SUCCESS; the file will be deleted once it is sent.

Note: I've also tested that the retry attempt happens again from starting after the 30-minute cool-off period.

## Breaking Changes

- Now, the SDK keeps on retrying the retryable error before proceeding to the next batch of events.

## Maintainers Checklist

- [ ] CI tests have passed.

image

### DIFF
--- a/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/network/EventUploadResult.kt
+++ b/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/network/EventUploadResult.kt
@@ -1,0 +1,81 @@
+package com.rudderstack.sdk.kotlin.core.internals.network
+
+import com.rudderstack.sdk.kotlin.core.internals.utils.Result
+
+/**
+ * `EventUploadResult` is a sealed interface representing the result of an event upload attempt.
+ * It can either be a success or an error.
+ */
+internal sealed interface EventUploadResult
+
+/**
+ * `EventUploadSuccess` represents a successful event upload.
+ * @property response The response from the server.
+ */
+internal data class EventUploadSuccess(val response: String) : EventUploadResult
+
+/**
+ * `EventUploadError` is a sealed interface representing an error that occurred during event upload.
+ * It can be either a retry able error or a non-retry able error.
+ */
+internal sealed interface EventUploadError : EventUploadResult
+
+/**
+ * `RetryAbleError` is a sealed interface representing an event upload error that can be retried.
+ */
+internal sealed interface RetryAbleError : EventUploadError
+
+/**
+ * `NonRetryAbleError` is a sealed interface representing an event upload error that cannot be retried.
+ */
+internal sealed interface NonRetryAbleError : EventUploadError
+
+/**
+ * `RetryAbleEventUploadError` is an enum class representing the different types of retryable event upload errors.
+ *  @property ERROR_RETRY A generic error that can be retried.
+ *  @property ERROR_NETWORK_UNAVAILABLE An error indicating that the network is unavailable, and the upload can be retried.
+ *  @property ERROR_UNKNOWN An unknown error occurred, and the upload can be retried.
+ */
+internal enum class RetryAbleEventUploadError : RetryAbleError {
+    ERROR_RETRY,
+    ERROR_NETWORK_UNAVAILABLE,
+    ERROR_UNKNOWN
+}
+
+/**
+ * `NonRetryAbleEventUploadError` is an enum class representing the different types of non-retryable event upload errors.
+ *  @property ERROR_400 An error indicating that the request was invalid (e.g., missing or malformed body).
+ *  @property ERROR_401 An error indicating that the request was unauthorized.
+ *  @property ERROR_404 An error indicating that the resource was not found (e.g., the source is disabled).
+ *  @property ERROR_413 An error indicating that the payload size exceeds the maximum allowed limit.
+ */
+internal enum class NonRetryAbleEventUploadError : NonRetryAbleError {
+    ERROR_400,
+    ERROR_401,
+    ERROR_404,
+    ERROR_413
+}
+
+/**
+ * Extension function to convert a `NetworkResult` to an `EventUploadResult`.
+ * @return An `EventUploadResult` representing the result of the network operation.
+ */
+internal fun NetworkResult.toEventUploadError(): EventUploadResult {
+    return when (this) {
+        is Result.Success -> {
+            return EventUploadSuccess(response)
+        }
+
+        is Result.Failure -> {
+            when (this.error) {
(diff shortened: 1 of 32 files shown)


core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/TestClock.kt
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * A clock implementation for testing that uses the TestCoroutineScheduler
+ * to provide consistent time values during tests.
+ */
+class TestClock(private val scheduler: TestCoroutineScheduler) {
+    /**
+     * Returns the current virtual time from the test scheduler in milliseconds.
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    fun currentTimeInMillis(): Long = scheduler.currentTime
```
ChryssaAliferi, line 16: 👍 Great stuff!!

ChryssaAliferi (approved):
This naming, `RetryAble` / `NonRetryAble` naming, is an internal convention mostly and should be bound to change in case the business logic of our SDK changes. For internal communication and visibility we will keep it, in general I would be in favor of more generic naming approaches. 

Amazing work!

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/network/EventUploadResult.kt
```diff
@@ -0,0 +1,81 @@
+package com.rudderstack.sdk.kotlin.core.internals.network
+
+import com.rudderstack.sdk.kotlin.core.internals.utils.Result
+
+/**
+ * `EventUploadResult` is a sealed interface representing the result of an event upload attempt.
+ * It can either be a success or an error.
+ */
+internal sealed interface EventUploadResult
```
vgupta98, line 9: Creating a whole new interface to handle results for event upload even though we already have a `Result` class and a `NetworkResult` is a bit of an overkill. 
It definitely improves readability in `EventUpload` class, but I will try once to use the existing `NetworkResult` class itself while maintaining the readability. If not possible, I will resolve this conversation.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/policies/backoff/MaxAttemptsExponentialBackoff.kt
```diff
@@ -0,0 +1,58 @@
...
+ * - Applies increasing delays between retry attempts
+ * - Tracks attempt count and enforces maximum limits
+ * - Suspends execution between attempts with appropriate delays
+ * - Implements a longer cool-off period when max attempts are reached
+ *
+ * @param maxAttempts Maximum retries before entering cool-off (default: 5)
+ * @param coolOffPeriod Duration in minutes (default: 30 minutes)
+ * @param exponentialBackOffPolicy Delay calculation policy
+ */
+internal class MaxAttemptsExponentialBackoff(
```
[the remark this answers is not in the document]
1abhishekpandey: Thanks for the comment — it helped me make the MaxAttempt class more generic. I’ll keep this in mind moving forward.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/network/EventUploadResult.kt
vgupta98: Let's go ahead with this approach for now.

vgupta98 (approved):

PR #173, opened 2025-05-12
fix: userId set as empty issue

vgupta98:
## Description

This PR fixes an issue in which empty `userId` was stored in the `Storage` even after calling the `identify` API. This led to missing `userId` in payloads of events even after the user was identified.

## Type of change

- [ ] Documentation update

## Implementation Details

Following changes are done:
1. The logic of clearing the `userId` and `traits` before setting them in the `SetUserIdAndTraitsAction` is removed. Now, when the `identify` call is made, the `userId` and `traits` are overwritten instead of first getting cleared and then written.
2. The `storageDispatcher` is refactored. There are two dispatchers used for storage updation now - `fileStorageDispatcher` and `keyValueStorageDispatcher`. These dispatchers are used for updating of batch files and updating of key-value storage respectively. Both are made as single threaded dispatchers.
3. The number of threads for both the dispatchers is set to 1 to ensure that the coroutines dispatched on it are executed in a sequential manner. This ensures the consistency of values stored in persistence when it is updated using a series of coroutines.
4. The tests in the `SetUserIdAndTraitsActionTest` are updated to remove the verification for clearing of storage.

## Checklist

## How to test?

Ensure that `userId` and `traits` are always present in the payloads after the `identify` call is made and until the `reset` call is made. Check this logic after even after the app is restarted.

## Maintainers Checklist

- [ ] CI tests have passed.

## Additional Context

The missing `userId` issue was occurring because of the following reason - there were two coroutines which were launched to update the `storage` when an `identify` call was made:
1. First coroutine cleared the old userId and traits from storage in the SetUserIdAndTraitsAction.
2. Second coroutine wrote the updated userId and traits on the storage.

Now, once the coroutines are dispatched, they can be run in any order (in case of a multi-threaded dispatcher) by the compiler. Randomly, the clearing coroutine (even though it was dispatched before) was running after the setting coroutine which lead to clearing of the userId and traits after they were being set.

### Performance analysis after increasing number of dispatchers
Since, the dispatcher is created using the method -> `Dispatchers.IO.limitedParallelism(1)`, even though it assigns a new thread for coroutines running on that dispatcher, the thread is assigned from a pool of reserved threads and is killed if not required for a long time. Hence, overall it does not increase the active thread count when the SDK is running but is idle.
In stress scenarios however, we want new threads to spawn to cater the increased demand for processing, hence it is the ideal behaviour.
Hence, overall, there is no performance impact of having a new single threaded dispatcher in the codebase.

### DIFF
--- a/android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/AnalyticsTest.kt
+++ b/android/src/test/kotlin/com/rudderstack/sdk/kotlin/android/AnalyticsTest.kt
@@ -93,6 +93,7 @@ class AnalyticsTest {
             every { analyticsScope } returns testScope
             every { analyticsDispatcher } returns testDispatcher
             every { fileStorageDispatcher } returns testDispatcher
+            every { keyValueStorageDispatcher } returns testDispatcher
             every { networkDispatcher } returns testDispatcher
             every { integrationsDispatcher } returns testDispatcher
             every { storage } returns mockStorage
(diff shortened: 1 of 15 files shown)


1abhishekpandey (approved):

ChryssaAliferi (approved):
Greaat!

PR #313, opened 2026-04-15
feat(sprig)!: add Sprig device-mode integration

vgupta98:
## Description
Add a new Sprig (UserLeap) device-mode integration for the RudderStack Kotlin SDK. This integration supports `identify`, `track`, and `reset` calls, mapping them to the corresponding Sprig SDK methods. It also handles activity lifecycle to enable in-app survey presentation via `trackAndPresent`.

## Type of change
- [ ] Documentation update

## Implementation Details
- **Module scaffold**: Created `integrations/sprig` module with `build.gradle.kts`, ProGuard rules, AndroidManifest, and publishing configuration. Registered `Sprig` in `RudderStackBuildConfig` and `settings.gradle.kts`.
- **Dependencies**: Added `userleap-android-sdk` (v2.23.0+) and `androidx-fragment-ktx` (v1.8.9) to the version catalog.
- **SprigConfig**: `@Serializable` data class for deserializing the `environmentId` from the destination config.
- **SprigIntegration**: Implements `StandardIntegration`, `IntegrationPlugin`, and `ActivityLifecycleObserver`.
  - `create()` — Initializes the Sprig SDK with the environment ID and registers lifecycle observer (idempotent).
  - `identify()` — Sets user identifier via `setUserIdentifier` and maps traits (email via `setEmailAddress`, other primitives via `setVisitorAttribute`).
  - `track()` — Builds an `EventPayload` and calls `trackAndPresent` when a `FragmentActivity` is available, otherwise falls back to `track`.
  - `reset()` — Calls `sprig.logout()`.
  - Activity lifecycle callbacks track the current `FragmentActivity` for survey presentation.
- **Utils**: Helper functions for config parsing (`parseConfig`), trait extraction (`IdentifyEvent.traits`), attribute mapping (`setSprigAttributes`), and `JsonObject.toStringMap()`. Attribute keys are validated (length < 256, no `!` prefix).
- **Unit tests**: Comprehensive tests in `SprigIntegrationTest` (create, identify, track, reset, lifecycle) and `UtilTest` (parseConfig, setSprigAttributes, toStringMap) using MockK.

## Checklist

## How to test?
1. Run unit tests: `./gradlew :integrations:sprig:test` — all tests should pass.
2. Integration test in a sample app:
   - Add the Sprig integration to a RudderStack-configured Android app.
   - Verify `identify` sets user ID and traits in the Sprig dashboard.
   - Verify `track` sends events; if a `FragmentActivity` is in the foreground, surveys should present.
   - Verify `reset` clears user state.

## Breaking Changes
None

## Maintainers Checklist
- [ ] CI tests have passed.

N/A

## Additional Context
- Sprig SDK is published under the `com.userleap` Maven group. The version range is `[2.23.0, 3.0.0)`.
- `FragmentActivity` dependency (`androidx-fragment-ktx`) is required because the Sprig SDK's `trackAndPresent` API needs a `FragmentActivity` to present in-app surveys.


integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/SprigIntegration.kt
```diff
@@ -0,0 +1,167 @@
...
+        if (activity == null) {
+            sprig.track(eventPayload)
+        } else {
+            activity.runOnUiThread {
+                if (activity.canHostSurvey()) {
+                    sprig.trackAndPresent(eventPayload, activity)
+                } else {
+                    sprig.track(eventPayload)
+                }
+            }
```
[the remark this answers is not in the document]
vgupta98: - The extra UI-thread hop only happens when currentActivity is non-null but canHostSurvey() is false. Per the documented contract, hosts call setFragmentActivity(null)
  in onPause, so currentActivity is null during background states and this path isn't exercised.                                                                          
- The inner canHostSurvey() check inside runOnUiThread is required to guard against a TOCTOU race — the activity can transition out of STARTED between the outer check
  and the UI-thread dispatch, and trackAndPresent on an activity that has saved its instance state throws IllegalStateException (see commit 10fa6d7). Hoisting the check  
  out means either (a) duplicating it or (b) losing the race guard.                               
- sprig.track() is a lightweight enqueue, so even when the edge case does fire (host forgot to clear the reference), the cost of the UI-thread hop is negligible.

integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/SprigIntegration.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    public override fun create(destinationConfig: JsonObject) {
+        sprig ?: run {
+            destinationConfig.parseConfig<SprigConfig>(analytics.logger)?.let { config ->
+                loggingListener = registerSprigLogging(analytics.logger, analytics.configuration.logLevel)
+                sprig = Sprig
+                sprig?.configure(analytics.application.applicationContext, config.environmentId)
+                (analytics as? AndroidAnalytics)?.addLifecycleObserver(this)
+                analytics.logger.info("SprigIntegration: Sprig SDK initialized")
+            }
```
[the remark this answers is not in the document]
vgupta98: 1. Not every FragmentActivity should host surveys. Splash screens, modals, background-work activities, or activities mid-transition are poor candidates. Auto-wiring on
  every onActivityResumed captures activities the host may not want used for presentation. Explicit opt-in lets hosts choose which activities are eligible (the pattern is
   standard for overlay/survey SDKs).                                                             
  2. The "longer-than-necessary reference" risk is bounded. onActivityDestroyed (still wired via the lifecycle observer) clears the reference as a safety net, so a       
  forgotten setFragmentActivity(null) in onPause at worst keeps the reference until destroy — no actual leak, since the activity would be GC-eligible anyway. The Sprig   
  integration also clears the reference in teardown().
  3. trackAndPresent is guarded. Inside track() we check activity.canHostSurvey() (!isFinishing && !isDestroyed && lifecycle ≥ STARTED) on the UI thread before           
  presenting, so a stale reference from a backgrounded activity falls back to Sprig.track() rather than attempting a fragment transaction.                                
  4. The contract is documented in the README (Presenting in-app surveys) and setFragmentActivity KDoc, including the rationale for clearing in onPause.

integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/SprigIntegration.kt
```diff
@@ -0,0 +1,167 @@
...
+     * state, [track] falls back to `Sprig.track` and in-app experiences are not presented.
+     */
+    fun setFragmentActivity(activity: FragmentActivity?) {
+        currentActivity = activity
+    }
+
+    override fun onActivityDestroyed(activity: Activity) {
+        if (activity == currentActivity) {
+            currentActivity = null
+        }
```
1abhishekpandey, line 131: Do we really need to check for `activity == currentActivity` before setting it to `null`?

integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/Utils.kt
```diff
@@ -0,0 +1,131 @@
...
+
+internal fun setSprigAttributes(sprig: Sprig, attributes: JsonObject, logger: Logger) {
+    attributes[EMAIL_KEY]?.let { element ->
+        (element as? JsonPrimitive)?.toStringOrNull()?.let { email ->
+            sprig.setEmailAddress(email)
+        }
+    }
+
+    for ((key, value) in attributes) {
+        if (key != EMAIL_KEY) {
```
1abhishekpandey: There could be more standard keys in future. Let's not check only for a specific key. Let's store it in an array and then check it.

integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/Utils.kt
```diff
@@ -0,0 +1,131 @@
...
+
+private fun setVisitorAttribute(sprig: Sprig, key: String, value: JsonElement, logger: Logger) {
+    if (key.length >= MAX_ATTRIBUTE_KEY_LENGTH || key.startsWith("!")) {
+        logger.warn(
+            "SprigIntegration: '$key' is not a valid property name. " +
+                "Property names must be less than $MAX_ATTRIBUTE_KEY_LENGTH characters " +
+                "and cannot start with '!'. Ignoring property"
+        )
+        return
+    }
```
1abhishekpandey: Instead of dropping for `key.length >= MAX_ATTRIBUTE_KEY_LENGTH` can we instead trim it and send?

integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/Utils.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    val primitive = value as? JsonPrimitive ?: run {
+        logger.warn(
+            "SprigIntegration: '$key' has an unsupported value type. " +
+                "Only String, Int, and Boolean are accepted. Ignoring property"
+        )
+        return
+    }
+
+    val handled = primitive.toStringOrNull()?.also { sprig.setVisitorAttribute(key, it) }
```
1abhishekpandey: Let's use a better variable name for `handled`. Currently, it does not seem to help with readability.

integrations/sprig/src/test/kotlin/com/rudderstack/integration/kotlin/sprig/SprigIntegrationTest.kt
```diff
@@ -0,0 +1,455 @@
...
+
+        every { mockAnalytics.application } returns mockApplication
+        every { mockApplication.applicationContext } returns mockApplicationContext
+
+        sprigIntegration = SprigIntegration().also {
+            it.analytics = mockAnalytics
+        }
+    }
+
+    // region create
```
1abhishekpandey: Let's use a nested class instead of a comment.

integrations/sprig/src/test/kotlin/com/rudderstack/integration/kotlin/sprig/UtilTest.kt
```diff
@@ -0,0 +1,188 @@
...
+    @BeforeEach
+    fun setup() {
+        mockkObject(Sprig)
+        every { Sprig.setEmailAddress(any()) } just Runs
+        every { Sprig.setVisitorAttribute(any(), any<String>()) } just Runs
+        every { Sprig.setVisitorAttribute(any(), any<Int>()) } just Runs
+        every { Sprig.setVisitorAttribute(any(), any<Boolean>()) } just Runs
+    }
+
+    // region parseConfig
```
1abhishekpandey: Let's use nested class instead of comments.

1abhishekpandey (commented):
Great work.

integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/SprigIntegration.kt
vgupta98: Yes, there can be multiple activities in an app. And, if an activity is getting destroyed, it might not necessarily be the same activity as the user set using `setFragmentActivity`.

1abhishekpandey: Can you add this as a comment?

integrations/sprig/src/main/kotlin/com/rudderstack/integration/kotlin/sprig/Utils.kt
vgupta98: I need more clarification for this - we are calling a particular API for the email key. If a new key comes in the future, we would be calling a new API for it. How will we benefit from having it stored in an array.

1abhishekpandey: Ove here: `if (key != EMAIL_KEY)` we are checking for a specific key. Suppose, in future, there's a `phone_no` key. As per the current implementation, you need to write the code like this: `if (key != EMAIL_KEY && key != PHONE_NO)` this line will keep on expanding for new keys. 

The better way would be to add such standard keys in an array/set and just check if the keys exist over there or not.

There are multiple other benefits of it e.g., better test cases and code maintainability.

vgupta98: I can move the keys to an array/set, but we still need checks like -> "if key is `email`, call `setEmailAddress`, if key is `phone`, call `setPhone` (example) etc.".

1abhishekpandey: Those are two different codes: the setting of standard and custom traits. 

I've only commented on the portion about setting the custom traits.

(I've intentionally skipped this) You may refactor the code to handle both things in different methods, as standard traits handling may evolve in future, whereas custom ones will not.

1abhishekpandey (approved):

PR #325, opened 2026-07-09
fix(device-mode): apply event filtering for late-subscribing plugins

vgupta98:
## Description

Client-side event filtering (deny/allow lists) for device-mode destinations silently stopped working on a **warm launch when the dashboard config was unchanged since the last run** — which is the common case for essentially every app launch after the first install. Denied events were forwarded to the destination anyway.

This is deterministic, not flaky, and the same underlying flaw left a latent risk in destination initialization (`IntegrationsManagementPlugin`), which happened to win the same timing "race" in practice.

**Root cause:** `EventFilteringPlugin` and `IntegrationsManagementPlugin` observed `sourceConfigState` via `dropInitialState()` (`drop(1)`), which skips the first emission **positionally**. That is only correct when the collector subscribes *before* the first real value is published. But a device-mode integration is added by app code **after** the `Analytics` constructor has already dispatched the **cached** source config, so the plugin subscribes late, its first observed value is the *real* cached config, and `drop(1)` discards it. The subsequent network refresh is identical, so `StateFlow` de-dupes it, and the deny list is never applied.

- Cold launch: works (seed is dropped; network arrives after subscribe).
- Warm launch, config changed: works (distinct network emission after subscribe).
- Warm launch, config unchanged: silently leaks. ← the bug.

The positional `dropInitialState()` was introduced in #143 and applied uniformly to all `sourceConfigState` consumers, without accounting for late subscribers.

## Type of change

- [ ] Documentation update

## Implementation Details

- **Replaced the positional skip with a content/lifecycle-based one.** Added `State.observeDispatched()`, which emits values only once a value has actually been **dispatched**, skipping just the constructor seed. It is correct regardless of subscription timing: a collector subscribing *before* the first dispatch skips the seed and then receives every dispatched value; a collector subscribing *after* the first dispatch immediately receives the current (already dispatched) value and every subsequent one.
- **Removed `dropInitialState()`** (`StateUtils.kt`) and migrated `EventFilteringPlugin` and `IntegrationsManagementPlugin` to `observeDispatched()`.
- **`State` now composes a `MutableStateFlow` instead of inheriting it.** `MutableStateFlow` is documented as *not stable for inheritance* in third-party libraries. `State` now exposes a read-only `flow: StateFlow<T>` for observation and a `value` accessor for synchronous reads; the only way to mutate is `dispatch()`, so there is no bypass of the reducer.
- **Routed the remaining flow consumers through `State.flow`** — `EventQueue` (flush scheduling) and `ConnectivityUtils` (connection-available gate). All `.value` reads and `dispatch()` calls are unchanged.
- **Added regression tests** for the previously untested late-subscriber ordering.

## Checklist

## How to test?

**Automated**

- `./gradlew :core:test :android:testDebugUnitTest` — full suites green (837 tests, 0 failures).
- `StateTest` — new cases prove `observeDispatched()` is timing-independent: an early subscriber skips the seed, and a **late** subscriber receives the current dispatched value (not the seed).
- `EventFilteringPluginLateSubscriberTest` — reproduces the exact warm-launch ordering (config dispatched *before* subscribe, then an identical refresh that `StateFlow` de-dupes) and asserts the blacklisted event is dropped.

**Manual (device-mode, e.g. Braze with a deny list)**

1. Configure a device-mode destination with a couple of denied events in the dashboard.
2. Cold launch (fresh install), fire the denied events + one allowed control event → denied events must be dropped, control forwarded.
3. Relaunch the app without changing the dashboard config (warm, unchanged) and repeat → denied events must still be dropped.
4. Change the deny list in the dashboard, relaunch, repeat → denied events must be dropped.

Validated on-device (Pixel 9a): the warm-unchanged case that leaked on 8/8 runs before the fix now drops the denied events on 6/6 consecutive relaunches; cold and warm-changed also correct.

## Breaking Changes

None for public APIs. `State` is an internal (`@InternalRudderApi`) type; it no longer implements `MutableStateFlow`, so internal consumers observe via `State.flow` and mutate via `dispatch()`. All in-tree call sites are updated in this PR.

## Maintainers Checklist

- [ ] CI tests have passed.

N/A

## Additional Context

- Regression origin: #143 ("refactor: update source config success response logic"), which added `dropInitialState()` to all `sourceConfigState` consumers. That change added a late-subscription guard for integrations (`isSourceEnabledFetchedAtLeastOnce`) but not for the plugins nested inside them.
- The switch to composition also resolves the `MutableStateFlow` "not stable for inheritance" concern and makes `dispatch()` the single source of truth for state mutation.


android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPlugin.kt
```diff
@@ -39,7 +38,7 @@ internal class IntegrationsManagementPlugin : Plugin {
         integrationPluginChain.analytics = analytics
         analytics.withIntegrationsDispatcher {
             analytics.sourceConfigState
-                .dropInitialState()
+                .observeDispatched()
```
dciccale, line 41: Could we add a regression test for this consumer as well? The existing tests subscribe before dispatching the config, while the case fixed here has the opposite ordering. Something along these lines:

```kotlin
@Test
fun `given config dispatched before setup, when an integration is added, then it is initialised`() =
    runTest {
        val state = State(SourceConfig.initialState())
        every { mockAnalytics.sourceConfigState } returns state

        state.dispatch(SourceConfig.UpdateAction(sourceConfigWithCorrectApiKey))

        integrationsManagementPlugin.setup(mockAnalytics)
        advanceUntilIdle()

        integrationsManagementPlugin.addIntegration(integrationPlugin)
        advanceUntilIdle()

        verify(exactly = 1) {
            integrationPlugin.initDestination(sourceConfigWithCorrectApiKey)
        }
    }
```

This would fail with the previous `dropInitialState()` implementation because the cached config would be discarded and the subsequently added integration would never initialize. It would also cover this path independently from the `EventFilteringPlugin` regression tests.

dciccale (changes requested):
check comment

dciccale (approved):
LGTM

SKannaniOS (approved):

PR #333, opened 2026-07-30
feat(appsflyer)!: update AppsFlyer Android SDK to v7

vgupta98:
## Description

Updates the AppsFlyer integration to support **AppsFlyer Android SDK v7**.

AppsFlyer released v7 in May 2026, but `gradle/libs.versions.toml` pinned the dependency to
`strictly = "[6.17.0, 7.0.0)"`, published in the `appsflyer:1.3.1`/`1.3.2` POM as `[6.17.0, 7.0.0)`.
Because it uses `strictly`, any app that moves to AppsFlyer v7 gets an outright **dependency
resolution failure** — not a silent downgrade. The two constraints are unsatisfiable together.

The integration source does require changes, because v7 relocated its public classes and removed the
email API:

| API used | 6.17.0 | 7.0.1 |
|---|---|---|
| `AppsFlyerLib.getInstance()` | `com.appsflyer` | `com.appsflyer` (unchanged) |
| `setCustomerUserId(String)` | ✅ | ✅ |
| `logEvent(Context, String, Map)` | ✅ | ✅ |
| `setUserEmails(vararg String)` | ✅ plaintext `user_emails` | ❌ **removed** → `setUserEmail(String)`, sent as SHA-256 `email_hashed` |
| `AFInAppEventType` | `com.appsflyer` | **moved** to `com.appsflyer.share` |
| `AFInAppEventParameterName` | `com.appsflyer` | **moved** to `com.appsflyer.share` |

Critically, the event name and parameter **constants are byte-identical** between 6.17.0 and 7.0.1 —
verified with `javap -constants` across all 29 `AFInAppEventType` and all 78
`AFInAppEventParameterName` entries. So event mapping semantics are completely unchanged; only the
import paths move.

### Why the floor is 7.0.1, not 7.0.0

AppsFlyer made a **breaking API change within the 7.x line, in a patch release**:

| | 7.0.0 | 7.0.1 |
|---|---|---|
| `setUserEmails(vararg String)` | ✅ | ❌ removed |
| `setUserEmails(EmailsCryptType, vararg)` | ✅ | ❌ removed |
| `setUserEmail(String)` | ❌ | ✅ added |
| `setUserPhone` / `setUserFirstName` / `setUserLastName` / `clearUserPii` | ❌ | ✅ added |

No single source can compile against the whole `[7.0.0, 8.0.0)` range. Compiling against the 7.0.1
API (as this PR does) would give anyone resolving 7.0.0 a runtime `NoSuchMethodError`, so the floor is
set to 7.0.1. AppsFlyer 7.0.0 is superseded and was only current between 2026-05-26 and 2026-07-23.

Note: AppsFlyer's [official v6→v7 migration guide](https://dev.appsflyer.com/hc/docs/migrate-android-sdk-to-v7)
still documents `setUserEmails` with `EmailsCryptType` moved to `com.appsflyer.share`. That is **out of
date** relative to the shipped 7.0.1 artifact — neither `setUserEmails` nor `EmailsCryptType` exists
anywhere in the 7.0.1 AAR. All findings here come from inspecting the actual artifacts, not the docs.

### Behavioural change: identify email is now hashed

Event mapping is unchanged, but the `identify` email reaches AppsFlyer differently:

| | v6 | v7 |
|---|---|---|
| Field | `user_emails` | `email_hashed` |
| Value | `{"plain_el_arr":["premium.user@example.com"]}` | `7d4f5d6b…a12c677` |
| Form | plaintext, on every subsequent event | SHA-256 |

Confirmed the hash is of the same address:
`sha256("premium.user@example.com") = 7d4f5d6b2c2991d24d477a261f885afd0fbf5836c9dba696483cca528a12c677`

A privacy improvement, but a customer-visible change: anything keyed on the plaintext
`user_emails` field will now see a hash. Documented in the README.

## Type of change
- [x] Documentation update

## Implementation Details

**`gradle/libs.versions.toml`**
- Bumped `appsflyer-android-sdk` from `{ strictly = "[6.17.0, 7.0.0)", prefer = "6.17.0" }` to
  `{ strictly = "[7.0.1, 8.0.0)", prefer = "7.0.1" }`.

**`Utils.kt`**
- Repointed `AFInAppEventType` and `AFInAppEventParameterName` imports to `com.appsflyer.share`.
  These two imports account for 37 of the 38 compile errors the bump produces; no other line in the
  file changes.

**`AppsFlyerIntegration.kt`**
- Replaced the removed `setUserEmails(it)` with `setUserEmail(it)` in `identify()`. The integration
  only ever passed a single email, so this is behaviourally equivalent.

**`AppsFlyerIntegrationTest.kt` / `UtilsTest.kt`**
- Same mechanical import move, plus `setUserEmails` → `setUserEmail` in the two MockK verifications
  and their test names. No assertions changed — the existing expected event names and parameter keys
  all still hold, which is the relevant signal that mapping behaviour is intact.

**`README.md`**
- Supported native version `6.17.0` → `7.x (>= 7.0.1, < 8.0.0)`, with an explanation of the 7.0.1 floor.
- Added a note directing apps still on AppsFlyer v6 to the `1.x` line, linking AppsFlyer's migration guide.
- Rewrote the Usage snippet for the v7 startup model: `start()` no longer takes a `Context` and must be
  called from inside a `registerSessionReadyListener` block. Also moved `setLogLevel` before `init` so
  initialization logging is actually emitted.
- Noted that `com.android.installreferrer:installreferrer` does **not** need to be added by consumers:
  AppsFlyer's v7 guide lists it as required, but this integration already declares it (see
  `integrations/appsflyer/build.gradle.kts:118`, asserted by the POM verification test), so it arrives
  transitively. Flagging it avoids a reader following both docs adding a conflicting version.
- Added an "Install referrer collection" section: in v6 the SDK collected referrer data automatically,
  in v7 it is **opt-in** via `collectDataFromLauncherActivity`. This is a silent attribution behaviour
  change for apps that don't add that call.
- Noted the API 21 floor and the new `com.google.android.play:integrity` transitive dependency.

**Deliberately not changed**
- `RudderStackBuildConfig.kt` version and `CHANGELOG.md` are **not** hand-edited — `draft-new-release.yml`
  generates both via `bump-versions.sh` / `generate-changelog.sh` from conventional commits. Verified:
  `./scripts/releases/detect-affected-modules.sh` outputs `appsflyer|major|1.3.2|2.0.0`, and no other
  module is affected (`libs.versions.toml` does not map to any module in `file_to_module`).

## Checklist
- [x] I have added tests that prove my fix is effective or that my feature works. — No new tests; this is a dependency bump with no new logic. The existing 60-test suite passing unmodified (apart from the mechanical rename) against v7 is the relevant signal, backed by the end-to-end run below.
- [x] I have ensured that my code follows the project's code style. — `detekt` clean.
- [x] I have checked for potential performance impacts and optimized if necessary. — None; no logic change.
- [x] I have checked the code for security issues. — No credentials committed; `gitleaks` clean.
- [x] I have updated the changelog (if required). — Not required; generated at release time.

## How to test?

**Automated**
```bash
./gradlew :integrations:appsflyer:build
# BUILD SUCCESSFUL — detekt + lint clean
# AppsFlyerIntegrationTest 20/20, UtilsTest 39/39, AppsFlyerPomVerificationTest 1/1 (60 total, 0 failures)
```

Resolution and generated POM:
```
com.appsflyer:af-android-sdk:{strictly [7.0.1, 8.0.0); prefer 7.0.1} -> 7.0.1
├── com.appsflyer:af-android-sdk-base:7.0.1
└── com.google.android.play:integrity:1.3.0

pom-default.xml: <version>[7.0.1, 8.0.0)</version>
```

`AppsFlyerPomVerificationTest` passes without a fixture update — its `normalizeVersions()` masks
`<version>` values, so the range change is tolerated by design.

**End-to-end verification: v6 vs v7 A/B comparison (completed)**

Ran the same 23-event sequence (identify → 16 track → 5 screen) twice on one device
(Pixel 9a emulator, Android 17, registered as an AppsFlyer test device), separated by
`customer_user_id`:

| | Integration | AppsFlyer SDK |
|---|---|---|
| Baseline | `appsflyer:1.3.2` (released == develop) | 6.17.0 |
| This branch | locally published | 7.0.1 |

Compared both the client-side payloads (from the AppsFlyer SDK's own outgoing request
logs) and the server-side raw-data CSV exports from the AppsFlyer dashboard.

**Result: all 23 event names and `event_value` payloads are byte-for-byte identical.**
That covers custom-event naming, every `af_*` constant, array-valued parameters,
nested objects and rich screen naming.

One apparent discrepancy resolved: `af_ad_click` was missing from the v6 dashboard
export. The raw logs show v6 *did* send it, with a payload identical to v7's, and
AppsFlyer's server returned `404 Not Found` for that one request — a transient
server-side drop, not a mapping difference. Non-200 responses: v6 = 1, v7 = 0.

The only intentional difference is the identify email field (see above).

## Breaking Changes

**This integration now requires AppsFlyer Android SDK v7 (>= 7.0.1).** It can no longer resolve
alongside AppsFlyer 6.x — and because the constraint uses `strictly`, resolution fails outright rather
than silently downgrading.

Apps pinned to AppsFlyer 6.x cannot take this version. Migration path:

1. Stay on the **1.x** line of this integration if the app is not ready for AppsFlyer v7, **or**
2. Migrate the app to AppsFlyer SDK v7 and adopt this version. Required host-app changes:
   - Call `start()` with no arguments, from inside a `registerSessionReadyListener` block —
     `start(Context)` was removed.
   - Add `com.android.installreferrer:installreferrer:2.2`; v7 removed the broadcast receivers that
     previously collected referrer data.
   - Call `collectDataFromLauncherActivity(Activity)` from the launcher activity's `onCreate` if the app
     relies on referrer attribution — this is opt-in in v7, where v6 did it automatically.
   - Import `AppsFlyerConversionListener` from `com.appsflyer.share` if passing a listener to `init`.
   - Note that v6 APIs `setCollectIMEI`, `setCollectOaid`, `setSharingFilter*`,
     `setCustomerIdAndLogSession`, `validateAndLogInAppPurchase`, `subscribeForDeepLink`,
     `performOnDeepLinking` and `waitForCustomerUserId` were all removed.
   - Be aware the `identify` email now reaches AppsFlyer as a SHA-256 `email_hashed`
     value rather than plaintext `user_emails`. No app-side change is needed, but
     downstream consumers of the plaintext field may need reconfiguring.
   - See AppsFlyer's [Android v6 → v7 migration guide](https://dev.appsflyer.com/hc/docs/migrate-android-sdk-to-v7).

Because a dependency major bump forces a migration on consumers, this releases as **`appsflyer:2.0.0`**,
keeping the `1.x` line available for AppsFlyer v6 users. The release tooling derives this automatically
from the `feat!` commit.

## Maintainers Checklist
- [ ] CI tests have passed.

N/A

## Additional Context

- Linear: [SDK-5128](https://linear.app/rudderstack/issue/SDK-5128/android-kotlin-update-appsflyer-integration)
- **New transitive dependency:** v7 splits into `af-android-sdk` + `af-android-sdk-base` and adds
  `com.google.android.play:integrity:1.3.0` (which pulls `play-services-basement`). Apps get a new
  Google Play dependency with APK-size and ProGuard implications they did not explicitly opt into —
  worth a release-note callout.
- **kotlin-stdlib is forced to 2.0.21** (v7's `api` dependency, up from 1.8.22 in 6.17.0), which upgrades
  stdlib across the whole consumer graph. This compiles cleanly with the repo's Kotlin 1.9.0 — no
  metadata incompatibility — but consumers on older Kotlin should be aware.
- **minSdk 19 → 21** in the AppsFlyer AAR. No impact here: `RudderStackBuildConfig.AndroidBuild.MIN_SDK`
  is already 21.
- **Dual 6.x/7.x support is not possible.** Widening the range to accept both would not work: the
  constant classes live in different packages across the two majors, so a single source cannot compile
  against both. Reflection would be the only route and is not worth it for this.
- **Residual risk:** AppsFlyer broke a public API in a patch release (7.0.0 → 7.0.1) once already, so the
  `[7.0.1, 8.0.0)` range carries some risk that a future 7.x patch breaks compilation again. Narrowing
  further would be unhelpfully rigid, but it is worth watching new 7.x releases.
- The POM validator baselines in `.claude/skills/rsa-validate-*-pom/references/` still record the old
  `[6.17.0, 7.0.0)` range. Those are post-release validation baselines and are intentionally left for the
  release-validation step rather than updated here.


dciccale (changes requested):
Please address the failing PR title check and clarify or remove the redundant install-referrer dependency instruction. The AppsFlyer v7 implementation and validation otherwise look good.

dciccale (approved):

PR #340, opened 2026-08-13
feat(core): add consent management configuration and setConsent API

SKannaniOS:
## Description
Adds consent management support to the Kotlin SDK. Users can now configure consent at SDK initialization via a new `ConsentManagementConfiguration` on `Configuration`, and update consent values at runtime through a new `Analytics.setConsent(ConsentManagementOptions)` API. Consent values (allowed/denied consent category IDs) are held in a Redux-style `State`, matching the SDK's existing state-management pattern, and full Java interop parity is provided via `JavaAnalytics` and builder classes. Consent management is disabled by default, so existing behavior is unchanged.

## Type of change
- [ ] Documentation update

## Implementation Details
- **`ConsentManagementConfiguration`** (`core/.../consent/ConsentManagementConfiguration.kt`): new load-time configuration with `enabled` (default `false`), `provider` (currently only `ConsentManagementProvider.CUSTOM`, wire value `"custom"`), and `allowedConsentIds` / `deniedConsentIds` lists. Wired into both core and Android `Configuration` classes as a new `consentManagement` property with a disabled default.
- **`ConsentManagementOptions`** (`core/.../consent/ConsentManagementOptions.kt`): runtime payload for `setConsent`. The lists fully replace the current consent state (not a delta); an empty options instance clears all consent state.
- **`ConsentManagementState`** (`core/.../internals/models/consent/ConsentManagementState.kt`): internal Redux-style state holding the current consent values. Consent IDs are trimmed and empties dropped via `normalized()`. An `initialized` flag stays `false` until non-empty consent values are supplied, keeping consent evaluation fail-open (no destination blocked) until then.
- **`SetConsentAction`** (`core/.../internals/models/consent/SetConsentAction.kt`): `StateAction` that replaces both consent lists and recomputes `initialized`. `enabled` and `provider` are load-time settings and are never modified at runtime, matching the JS SDK behavior.
- **`Analytics.setConsent(options)`** (`core/.../Analytics.kt`): new public API that dispatches `SetConsentAction` onto the internal `consentManagementState`. It is a no-op (with a warning log) when consent management is disabled in configuration, and respects the analytics-shutdown guard.
- **Java parity**: `JavaAnalytics.setConsent(options)`, a new `ConsentManagementConfigurationBuilder`, and `setConsentManagement(...)` on both the core and Android `ConfigurationBuilder` classes.
- **Tests**: new coverage for the configuration model, state initialization/normalization, the `SetConsentAction` reducer, the `setConsent` API (including disabled and shutdown paths), and all Java-compat builders/APIs (~570 lines of test code across core and android modules).

## Checklist

## How to test?
- Run the core unit tests: `./gradlew :core:test --tests "*Consent*"` (covers `ConsentManagementConfigurationTest`, `SetConsentApiTest`, `ConsentManagementStateTest`, `SetConsentActionTest`, `ConsentManagementConfigurationBuilderTest`, and the updated `ConfigurationBuilderTest` / `JavaAnalyticsTest`).
- Run the Android unit tests: `./gradlew :android:testDebugUnitTest` (covers the Android `ConfigurationTest` and `ConfigurationBuilderTest` additions).
- Manual verification: initialize the SDK with `consentManagement = ConsentManagementConfiguration(enabled = true, allowedConsentIds = listOf(...))`, then call `analytics.setConsent(ConsentManagementOptions(...))` and confirm the state updates; with consent management disabled, confirm `setConsent` logs a warning and has no effect.

## Breaking Changes
None. `consentManagement` is a new optional constructor parameter with a default value on `Configuration` (core and Android), `setConsent` is a new API, and consent management is disabled by default.

## Maintainers Checklist
- [ ] CI tests have passed.

No UI changes in this PR.

## Additional Context
- This PR introduces the consent configuration and state model only; event stamping (`context.consentManagement`) and destination gating based on consent are expected in follow-up work — while uninitialized, consent evaluation fails open and no destination is blocked.
- The full-replacement (non-merge) semantics of `setConsent` intentionally match the RudderStack JS SDK.


dciccale (approved):

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/Analytics.kt
```diff
@@ -411,6 +421,33 @@ open class Analytics protected constructor(
...
+     * The supplied lists fully replace the existing consent state — callers always pass the
+     * complete current state, not a delta. Passing an empty [ConsentManagementOptions] clears
+     * both lists and reverts consent evaluation to the uninitialized fail-open state.
+     *
+     * This method has no effect while consent management is disabled in [Configuration];
+     * enabling consent management is a load-time decision.
+     *
+     * @param options The consent values to apply.
+     */
+    open fun setConsent(options: ConsentManagementOptions) {
```
vgupta98: Why have kept it this API as `open`? Are we going to override it in Android Analytics?

SKannaniOS: Every public API is `open` in this class for this API also `open` for consistency. There's no android override for `setConsent` in this milestone.

vgupta98: We can remove `open` for now. Only the APIs which are overriden are actually open.

vgupta98 (approved):

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/models/consent/SetConsentAction.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    override fun reduce(currentState: ConsentManagementState): ConsentManagementState {
+        if (!currentState.enabled) return currentState
+
+        val allowed = options.allowedConsentIds.normalized()
+        val denied = options.deniedConsentIds.normalized()
+        return currentState.copy(
+            allowedConsentIds = allowed,
+            deniedConsentIds = denied,
+            initialized = allowed.isNotEmpty() || denied.isNotEmpty(),
```
vgupta98: `setConsent(ConsentManagementOptions())` unblocks every destination.

A user tapping "Reject all" naturally produces empty lists, which sets `initialized = false`. `ConsentResolver` (SDK-5261) then returns consented for everything:

```kotlin
if (!state.enabled || !state.initialized) return true
```

So "the user refused everything" and "the user has not been asked yet" collapse into the same state, and both fail open. The intuitive call is the unsafe one.

The uninitialized fail-open default is right on app start — the issue is only that an explicit empty `setConsent` is indistinguishable from it.

If this is deliberate JS SDK parity, can we make it explicit in the public KDoc and on the docs site, that callers must express "deny all" as a non-empty `deniedConsentIds`? Otherwise, consider tracking "`setConsent` was called" separately from "consent values are non-empty".

SKannaniOS: Fixed.

The rule now: the two lists are never both empty while consent management is on.

- Empty setConsent is refused — warns, state unchanged.
- "Reject all" is a non-empty deniedConsentIds.
- initialized is gone. Enabled at load with neither list is a misconfiguration: consent management stays off for the session, with an info log.

So "refused everything" and "not asked yet" are no longer the same state.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/Analytics.kt
```diff
@@ -411,6 +429,45 @@ open class Analytics protected constructor(
...
+
+        val allowed = options.allowedConsentIds.normalized()
+        val denied = options.deniedConsentIds.normalized()
+        if (allowed.isEmpty() && denied.isEmpty()) {
+            logger.warn(
+                "Analytics(core): setConsent requires at least one consent ID; the call has no effect. " +
+                    "To deny every category, pass them in deniedConsentIds."
+            )
+            return
+        }
```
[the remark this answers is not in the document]
SKannaniOS: PR description updated.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/ConsentManagementState.kt
```diff
@@ -0,0 +1,52 @@
...
+            val allowed = configuration.allowedConsentIds.normalized()
+            val denied = configuration.deniedConsentIds.normalized()
+            val active = configuration.enabled && (allowed.isNotEmpty() || denied.isNotEmpty())
+            return ConsentManagementState(
+                enabled = active,
+                provider = configuration.provider,
+                allowedConsentIds = allowed,
+                deniedConsentIds = denied,
+            )
+        }
```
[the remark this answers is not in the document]
SKannaniOS: PR description updated.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/consent/ConsentManagementOptions.kt
```diff
@@ -0,0 +1,18 @@
...
+ * The supplied lists fully replace the current consent state — callers always pass the
+ * complete current state, not a delta. An omitted list defaults to empty and clears the
+ * corresponding values; an instance with both lists empty is rejected and leaves the
+ * current consent state unchanged. Updates apply only while consent management is
+ * enabled; otherwise the existing state is preserved.
+ *
+ * @param allowedConsentIds Consent category IDs the user has granted. Defaults to an empty list.
+ * @param deniedConsentIds Consent category IDs the user has denied. Defaults to an empty list.
+ */
+class ConsentManagementOptions @JvmOverloads constructor(
```
vgupta98: Two consistency points on this class:

1. It is not a `data class`, but `ConsentManagementConfiguration` is. So this type has no `equals`, `hashCode` or `toString`. Two option objects with the same IDs compare unequal.
2. There is no `ConsentManagementOptionsBuilder`, but `ConsentManagementConfigurationBuilder` exists in this PR. Java callers get a builder for one type and a constructor for the other.

Making it a data class and adding a builder would match the rest of the module.

vgupta98 (commented):
Reviewed this PR on its own diff against `develop`. The shape looks good — the inactive-when-no-IDs rule and the redacted `toString()` are both nice. Three small consistency points inline.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/ConsentManagementState.kt
```diff
@@ -0,0 +1,52 @@
...
+         * Consent IDs are trimmed and empties dropped. Enabling consent management without
+         * supplying either list is a configuration error: the state is built inactive, so the
+         * feature behaves exactly as if it had never been enabled.
+         */
+        fun initialState(configuration: ConsentManagementConfiguration): ConsentManagementState {
+            val allowed = configuration.allowedConsentIds.normalized()
+            val denied = configuration.deniedConsentIds.normalized()
+            val active = configuration.enabled && (allowed.isNotEmpty() || denied.isNotEmpty())
+            return ConsentManagementState(
+                enabled = active,
```
vgupta98: `enabled` here means "enabled **and** at least one consent ID was supplied". `ConsentManagementConfiguration.enabled` means "the customer asked for it". Same name, two different meanings.

The local variable on line 33 is already called `active`, which reads clearer. Renaming the state property to `active` would stop anyone comparing the two `enabled` flags and expecting them to agree.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/SetConsentAction.kt
```diff
@@ -0,0 +1,30 @@
...
+internal class SetConsentAction(
+    private val options: ConsentManagementOptions,
+) : ConsentManagementState.ConsentManagementStateAction {
+
+    override fun reduce(currentState: ConsentManagementState): ConsentManagementState {
+        if (!currentState.enabled) return currentState
+
+        val allowed = options.allowedConsentIds.normalized()
+        val denied = options.deniedConsentIds.normalized()
+        if (allowed.isEmpty() && denied.isEmpty()) return currentState
```
vgupta98: This empty check is duplicated. `Analytics.setConsent` already normalizes both lists and returns early when both are empty (`Analytics.kt:160-166`), so by the time the action runs the case cannot happen.

Two copies of the same rule can drift. Suggest keeping the validation in `setConsent` (it owns the warning log) and dropping it here, or the reverse — but not both.

vgupta98 (approved):

vgupta98 (approved):

PR #341, opened 2026-08-14
feat(core): add consent management stamping plugin

SKannaniOS:
## Description
Adds a `ConsentManagementPlugin` to the core event pipeline that stamps the current consent state into `context.consentManagement` on every event. This builds on the consent configuration and `setConsent` API introduced in SDK-5257 (base branch of this PR) and makes the consent values actually flow out on the event payload — provider, allowed consent IDs, and denied consent IDs. While consent management is disabled (the default), events pass through untouched, so any legacy `customContext` injection of the key keeps working unchanged.

## Type of change
- [ ] Documentation update

## Implementation Details
- **`ConsentManagementPlugin`** (`core/.../plugins/ConsentManagementPlugin.kt`): new `PreProcess` plugin. On each intercepted event it reads `analytics.consentManagementState.value`; if consent management is enabled it merges a complete `consentManagement` block (`provider`, `allowedConsentIds`, `deniedConsentIds`) into the event context with higher priority via `mergeWithHigherPriorityTo`, so the SDK-owned block always replaces any caller-injected value. If disabled, the event is returned untouched.
- **Legacy-key handling**: when the event context already contains a `consentManagement` key while consent management is enabled, the plugin logs a warning (without echoing the caller's consent IDs) advising migration to `setConsent()`, then overwrites the key. While disabled, a legacy injected key is preserved with no warning.
- **Registration** (`core/.../Analytics.kt`): the plugin is added in `setup()` between `LibraryInfoPlugin` and `RudderStackDataplanePlugin`, so stamping happens at event creation time — events already in the pipeline are not restamped.
- **Tests** (`core/.../plugins/ConsentManagementPluginTest.kt`): covers the exact stamped payload when enabled, the untouched pass-through when disabled, legacy-key replacement with warning (enabled) vs. preservation without warning (disabled), and that a `setConsent` state update between two events is reflected only on the later event.

## Checklist

## How to test?
- Run the new plugin tests: `./gradlew :core:test --tests "com.rudderstack.sdk.kotlin.core.plugins.ConsentManagementPluginTest"` (all 5 test cases).
- Run the full core suite to confirm no pipeline regressions: `./gradlew :core:test`.
- Manual verification: initialize the SDK with consent management enabled and consent IDs set, fire a `track` event, and confirm the outgoing payload contains `context.consentManagement` with `provider`, `allowedConsentIds`, and `deniedConsentIds`; with consent management disabled, confirm the key is absent (or preserved as-is when injected via custom context).

## Breaking Changes
None. The plugin is a no-op while consent management is disabled, which is the default. The only behavioral change for enabled users is intentional: the SDK-owned `consentManagement` block replaces any value injected into the event context manually, with a warning logged.

## Maintainers Checklist
- [ ] CI tests have passed.

No UI changes in this PR.

## Additional Context
- Stacked PR: targets `feat/sdk-5257-consent-configuration-and-setconsent-api`, which introduces `ConsentManagementConfiguration`, `ConsentManagementState`, and the `setConsent` API this plugin consumes. Merge that PR first.
- The stamp reflects the consent state at event creation; destination gating based on consent is not part of this PR.


vgupta98 (approved):
LGTM. The plugin is a clean no-op while consent management is disabled, and the stamp correctly takes priority over a caller-injected block.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/utils/AnalyticsUtils.kt
```diff
@@ -44,3 +46,15 @@ internal fun Analytics.runOnAnalyticsThreadAfter(previousJob: Job?, block: suspe
...
+
+/**
+ * The consent state of the android SDK instance, or an inactive state when there is none.
+ *
+ * A plugin's `analytics` is typed as core's [Analytics], which does not carry consent. An instance
+ * that is not the android one has no consent configured, so every consent check fails open - the
+ * same default the resolver applies to missing configuration.
+ */
+internal val Analytics.consentState: State<ConsentManagementState>
+    get() = (this as? AndroidAnalytics)?.consentManagementState ?: INACTIVE_CONSENT_STATE
```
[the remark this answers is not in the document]
SKannaniOS: `INACTIVE_CONSENT_STATE` is a read-only sentinel. It's private to this file and the accessor is internal, and every call site reads .value or collects the flow — nothing dispatches through it. `setConsent` dispatches against `AndroidAnalytics.consentManagementState` directly.

Returning a fresh State per access would allocate on the per-event path (`ConsentGatePlugin`, `IntegrationPlugin` — once per event per destination) to harden a branch that isn't reachable in production: every caller is an android-module plugin, so the cast always succeeds. The shared instance is deliberate — it keeps the fail-open decision in one place rather than repeating it at five call sites.

vgupta98 (commented):
Reviewed this PR on its own diff against `feat/sdk-5257-consent-configuration-and-setconsent-api`. The `consentState` extension is a clean way to reach android state from a core-typed plugin. Two notes inline.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/utils/AnalyticsUtils.kt
```diff
@@ -44,3 +46,15 @@ internal fun Analytics.runOnAnalyticsThreadAfter(previousJob: Job?, block: suspe
 internal fun AndroidAnalytics.runOnMainThread(block: suspend () -> Unit) = analyticsScope.launch(MAIN_DISPATCHER) {
     block()
 }
+
+private val INACTIVE_CONSENT_STATE: State<ConsentManagementState> = State(ConsentManagementState())
```
vgupta98, line 56: `INACTIVE_CONSENT_STATE` is one process-wide `State` shared by every non-android `Analytics` instance. `State` is mutable — it exposes `dispatch`.

Nothing dispatches to it today, so this is safe right now. But if anything ever does, the change leaks into every other instance that falls back to it. Worth either a comment saying it must never be dispatched to, or returning a fresh inactive state instead of a shared one.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/ConsentManagementPlugin.kt
```diff
@@ -0,0 +1,60 @@
...
+    override val pluginType: Plugin.PluginType = Plugin.PluginType.PreProcess
+
+    override lateinit var analytics: Analytics
+
+    override suspend fun intercept(event: Event): Event {
+        val state = analytics.consentState.value
+        if (!state.enabled) return event
+
+        if (event.context.containsKey(CONSENT_MANAGEMENT_KEY)) {
+            analytics.logger.warn(
```
vgupta98, line 50: This warning fires on **every event**, not once. A customer still injecting `consentManagement` through `customContext` gets one warn line per event for the life of the session.

Since the message is the same every time, consider warning once per session (a flag on the plugin) and keeping the per-event path silent.

vgupta98 (approved):

vgupta98 (approved):

PR #343, opened 2026-08-18
feat(android): gate device-mode destinations on user consent

SKannaniOS:
## Description
Device-mode destinations currently initialize and receive events regardless of the consent values supplied through `ConsentManagementConfiguration` / `setConsent`. This PR builds on the consent-management plugin work (SDK-5259) and adds the actual gating (SDK-5261): destinations whose dashboard `consentManagement` config is not satisfied by the current consent state are neither initialized nor delivered events, and a later consent change re-evaluates every destination in place. Cloud-mode delivery is untouched — the consent stamping already done on events continues, and gated events still flow to the dataplane.

Consent resolution is centralized in a single fail-open predicate (`ConsentResolver`) shared by both the initialization gate and the per-event gate, so unexpected or malformed config never blocks event delivery. The source config now also parses the root-level `consentManagementMetadata` block tolerantly, so a malformed value degrades to `null` instead of failing the whole source config parse.

## Type of change
- [ ] Documentation update

## Implementation Details
- **`core` — `SourceConfig.kt`**: added `consentManagementMetadata: ConsentManagementMetadata?` (with `ConsentProviderEntry { provider, resolutionStrategy }`) decoded from the root of the sourceConfig response. Uses a new `LenientConsentManagementMetadataSerializer` that swallows `SerializationException` / `IllegalArgumentException` and yields `null` rather than breaking the parse.
- **`core` — `ConsentResolver.kt` (new)**: stateless `ConsentResolver.resolve(state, destinationConfig)` implementing the rules: disabled or uninitialized state → consented; no entry matching the active provider → consented; empty/blank configured consent IDs → consented; otherwise match trimmed IDs against `allowedConsentIds` using `ConsentResolutionStrategy` (`and`/`all` → ALL, `or`/`any` → ANY, anything else → ALL). `deniedConsentIds` are never consulted. Any non-object element in a JSON list voids the whole list; non-string primitives read as `null`.
- **`core` — `ConsentManagementState.kt` / `Analytics.kt`**: promoted `ConsentManagementState` and `Analytics.consentManagementState` from `internal` to `@InternalRudderApi` public so the `:android` module can read them.
- **`android` — `IntegrationPlugin.kt`**: `findAndInitDestination` now checks `ConsentResolver` after the "not found" / "disabled" checks and skips `create()` when denied, notifying the failure callback with a new `ConsentDeniedException`. Refactored the three failure paths into a shared `notifyDestinationFailure` helper. `ConsentGatePlugin` is now installed as the first default sub-plugin, ahead of `EventFilteringPlugin` and `IntegrationOptionsPlugin`.
- **`android` — `ConsentGatePlugin.kt` (new)**: `PreProcess` plugin in each integration's private chain that reads the consent state live per event and drops the event (returns `null`) when the destination is denied. Destination config is tracked from `sourceConfigState`; evaluation fails open when there is no config.
- **`android` — `IntegrationsManagementPlugin.kt`**: the source-config observer is now `combine`d with `consentManagementState` (`onStart` emitting the current value, `distinctUntilChanged`) so any consent change triggers `findAndInitDestination` again on every registered `IntegrationPlugin` — late-creating destinations that become allowed and re-evaluating those that become denied.
- **`android` — `ConsentDeniedException.kt` (new)**: dedicated exception type surfaced through the destination-ready callback so callers can distinguish consent denial from other init failures.
- **Tests**: `ConsentResolverTest` (core), `SourceConfigConsentParsingTest` (core), `ConsentGatingTest` (android); `SourceConfigTest` / `TestSourceConfigProvider` updated so fixtures carry the root-level consent metadata; `mockAnalytics` test util now stubs `consentManagementState`.

## Checklist

## How to test?
Automated:
```bash
./gradlew :core:test --tests "*ConsentResolverTest" --tests "*SourceConfigConsentParsingTest" --tests "*SourceConfigTest"
./gradlew :android:testDebugUnitTest --tests "*ConsentGatingTest"
./gradlew detekt
```
Coverage added:
- `ConsentResolverTest` (core): every fail-open path (disabled, uninitialized, null config, no `consentManagement` array, no provider match, case-sensitive provider match, duplicate entries → first wins, empty/blank IDs), whitespace trimming, `and`/`all`/`or`/`any` strategy aliases and unknown-strategy default, denied IDs ignored, and malformed shapes (string instead of array, non-object list element, numeric provider / strategy).
- `SourceConfigConsentParsingTest` (core): metadata parsed from the response root, absent → `null`, unknown provider with extra fields preserved, missing fields default to `null`, malformed metadata → `null` with the rest of the config intact, initial state, encode/decode round trip, and metadata preserved through a disable-source reducer.
- `ConsentGatingTest` (android): denied destination skips `create()` and reports `ConsentDeniedException`; consent-management-disabled leaves behaviour unchanged; grant-after-denial late-creates the destination; events sent while denied are never delivered after a later grant; revoke mid-session stops delivery; a consent flip without reinitialization is dropped live by the event gate; denied events still pass through the outer chain unchanged for cloud delivery; post-grant events arrive in order.

Manual (sample app):
1. Configure a device-mode destination in the dashboard with a `consentManagement` entry for a provider (e.g. `oneTrust`) and one or more consent IDs.
2. Initialize the SDK with `ConsentManagementConfiguration` for that provider but without the required IDs in `allowedConsentIds` — the destination should not initialize and the ready callback should receive `ConsentDeniedException`; `track` calls should not reach the vendor SDK but should still be sent to the dataplane.
3. Call `setConsent` granting the required IDs — the destination should initialize and subsequent events should be delivered to it.
4. Revoke the IDs again — further events should be dropped for that destination.

## Breaking Changes
None. `ConsentManagementState` and `Analytics.consentManagementState` are exposed only under `@InternalRudderApi`; the public API surface is unchanged. Existing behaviour is preserved when consent management is disabled or when a destination has no `consentManagement` config.

## Maintainers Checklist
- [ ] CI tests have passed.

No UI changes in this PR.

## Additional Context
- Base branch: `feat/sdk-5259-consent-management-plugin` (consent configuration, `setConsent`, and event stamping). This PR should be merged after that one.
- The `ConsentGatePlugin` is placed first in each integration's default sub-plugin chain so denied events are dropped before event filtering / integration-options processing runs.
- Consent gating operates only on device-mode destinations; cloud-mode filtering by consent is handled server-side using the consent IDs already stamped on the event context.


vgupta98:
`ConsentDeniedException` never reaches the callback for Facebook.

The consent-denied path pushes `emptyJsonObject` through `update()` before it reports the reason:

```kotlin
safelyExecute(
    block = { if (isStandardIntegration) { update(destinationConfig); block() } },
    onException = { e -> isDestinationReady = false; notifyCallbacks(Result.Failure(e)) }
)
```

`FacebookDestinationConfig.appId` is required, and `facebook/Utils.kt:47` uses a `parseConfig()` with no null-safety, so `update(emptyJsonObject)` throws `MissingFieldException`. `block()` never runs, so the `notifyCallbacks(Result.Failure(ConsentDeniedException))` inside it is skipped — the caller gets `Field 'appID' is required` instead.

Gating itself still works, since `isDestinationReady = false` happens on both paths. Only the reported reason is wrong, which defeats the point of adding the new exception type.

Adjust and Braze are unaffected — their `parseConfig(logger)` returns null on empty input.

Suggest skipping `update()` on the denied path instead of pushing an empty config through it.

Minor, same root cause: AppsFlyer's `extractConfiguration({})` silently resets `isNewScreenEnabled` to false. It self-heals on re-consent.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/models/SourceConfig.kt
```diff
@@ -4,22 +4,33 @@ import com.rudderstack.sdk.kotlin.core.internals.statemanagement.StateAction
...
  * @property source The configuration details of a RudderStack source.
+ * @property consentManagementMetadata The consent provider metadata for the source, if configured.
  */
 @InternalRudderApi
 @Serializable
 data class SourceConfig(
-    val source: RudderServerConfigSource
+    val source: RudderServerConfigSource,
+    @Serializable(with = LenientConsentManagementMetadataSerializer::class)
+    val consentManagementMetadata: ConsentManagementMetadata? = null
```
vgupta98, line 33: `consentManagementMetadata` is parsed but never read.

`ConsentResolver` takes `resolutionStrategy` from the destination's own `consentManagement` entry. Nothing reads `SourceConfig.consentManagementMetadata` — on this branch only the model, `LenientConsentManagementMetadataSerializer` and `SourceConfigConsentParsingTest` reference it.

Is the source-level `resolutionStrategy` meant to be the fallback when a destination entry doesn't specify one? If so, that wire-up is missing.

If nothing is going to use it yet, can we drop it and add it back when something does? Right now it's about 60 lines plus a custom serializer that do nothing.

vgupta98:
Step 2 of the manual test isn't possible as written.

It says to initialize the SDK with `ConsentManagementConfiguration` for the `oneTrust` provider from step 1. `ConsentManagementProvider` only defines `CUSTOM("custom")`, so there's no way to do that.

Leaving it as `CUSTOM` doesn't work either — `ConsentResolver` only matches an entry whose provider equals the active one:

```kotlin
val entry = entries.firstOrNull {
    it.stringValue(PROVIDER_KEY) == state.provider.value
} ?: return true   // no match -> consented
```

With `oneTrust` in the dashboard nothing matches, so the destination initializes normally and no `ConsentDeniedException` is raised. Anyone following these steps would think gating is broken.

The unit tests all use `"custom"`, so the code is fine — the steps just need `custom` throughout.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPlugin.kt
```diff
@@ -103,24 +105,37 @@ abstract class IntegrationPlugin : EventPlugin {
             analytics.logger.debug("IntegrationPlugin[$key]: Non-standard integration, using empty config")
             return emptyJsonObject
```
vgupta98, line 109: Are custom (non-standard) integrations meant to be out of scope for consent enforcement?

Right now they are. `isDestinationConfigured` returns early here, before the consent check, so the init gate never runs. `ConsentGatePlugin` then looks the destination up by key in the source config, doesn't find it, and fails open — so the event gate doesn't run either. Events are still stamped with `context.consentManagement`; they just aren't blocked.

That may well be the right answer, since consent rules live in the dashboard destination config and a custom integration has none.

If so, can we state it in the docs? Someone enabling consent management will otherwise assume their custom integration is covered. Worth noting that `ConsentResolver` and `consentManagementState` are both public after this PR, so a customer could build their own gate — though both are marked `@InternalRudderApi`.

core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/models/SourceConfig.kt
SKannaniOS: You're right that nothing reads it, and nothing will until named providers land.

Keeping it anyway: the PRD explicitly asks the SDKs to parse it defensively in v1, so the named-provider work later is purely additive. JS does the same — it reads the block, then discards the value when the provider is custom.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationPlugin.kt
SKannaniOS: Right, and intentional.

vgupta98 (approved):
LGTM. Thanks for the fixes — the failure path no longer pushes an empty config into the destination, the manual test steps now use `custom`, the disabled source config is dropped before re-evaluation, and the consent gate collector is cancelled on teardown.

The rationale for keeping `consentManagementMetadata` parsing (PRD asks for defensive parsing in v1, JS does the same) is fair.

One follow-up, not blocking: please note somewhere in the docs that consent gating covers RudderStack-maintained integrations only, since custom integrations are intentionally out of scope.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/ConsentResolver.kt
```diff
@@ -0,0 +1,80 @@
...
+     * @param destinationConfig The destination's raw config, carrying optional `consentManagement` entries.
+     * @return `true` when the destination may receive events.
+     */
+    fun resolve(state: ConsentManagementState, destinationConfig: JsonObject?): Boolean {
+        // Rule 1: disabled -> consented.
+        if (!state.enabled) return true
+
+        // Rule 2: first entry matching the active provider; none (or no array) -> consented.
+        val entries = destinationConfig?.get(CONSENT_MANAGEMENT_KEY).asObjectList() ?: emptyList()
+        val entry = entries.firstOrNull { it.stringValue(PROVIDER_KEY) == state.provider.value } ?: return true
```
vgupta98, line 57: A destination configured on the dashboard for `oneTrust` or `ketch` never matches `"custom"`, so this returns `true` and every event is delivered.

That is correct — only `CUSTOM` is supported — but it is completely silent. A customer who set up OneTrust would see no gating and no clue why.

Suggest logging once per destination when it has `consentManagement` entries but none match the active provider. The `consentManagementMetadata` added to `SourceConfig` in this same PR carries exactly the provider list needed for that.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/ConsentResolver.kt
```diff
@@ -0,0 +1,80 @@
+package com.rudderstack.sdk.kotlin.android.models.consent
+
+import com.rudderstack.sdk.kotlin.android.models.consent.ConsentManagementState.Companion.normalized
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+
+private const val CONSENT_MANAGEMENT_KEY = "consentManagement"
```
vgupta98, line 8: `CONSENT_MANAGEMENT_KEY` and `PROVIDER_KEY` are also declared in `ConsentManagementPlugin.kt` (added in #341). Same strings, two files.

If the wire format ever changes, both need editing. Worth moving them to one shared internal object.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/ConsentResolver.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        // Rules 4 & 5: normalize the strategy, match against the allowed IDs only.
+        return when (ConsentResolutionStrategy.from(entry.stringValue(RESOLUTION_STRATEGY_KEY))) {
+            ConsentResolutionStrategy.ALL -> cleanedIds.all { it in state.allowedConsentIds }
+            ConsentResolutionStrategy.ANY -> cleanedIds.any { it in state.allowedConsentIds }
+        }
+    }
+
+    // Any non-object element voids the whole list, keeping list handling wholesale rather than per-entry.
+    private fun Any?.asObjectList(): List<JsonObject>? {
```
vgupta98, line 74: Checking intent here rather than flagging a bug.

One non-object element voids the **whole** list, so the destination falls open and receives events. On a privacy control, failing open because of one malformed array entry is a strong choice.

Is this deliberate, and does it match what the JS SDK does? If so, a line in the KDoc saying "one bad entry disables gating for this destination" would make the trade-off visible to whoever reads this next.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/plugins/devicemode/IntegrationsManagementPlugin.kt
```diff
@@ -37,8 +41,16 @@ internal class IntegrationsManagementPlugin : Plugin {
 
         integrationPluginChain.analytics = analytics
         analytics.withIntegrationsDispatcher {
-            analytics.sourceConfigState
-                .observeDispatched()
+            combine(
+                analytics.sourceConfigState.observeDispatched(),
+                analytics.consentState
```
vgupta98: These three operators are exactly `analytics.consentState.flow`.

`observeDispatched()` skips the seed on purpose, `onStart` puts it back, and `distinctUntilChanged` cleans up the duplicate that creates. `State.flow` already means "current value, then every change".

They are equivalent because `backing` only ever changes through `dispatch`, and `dispatch` also flips `dispatched` — so nothing can change between collection start and the first dispatch. All three cases match: value changes, dispatch reduces to the seed (StateFlow drops equal values), and collection starting after a dispatch.

The short form is also cheaper. `observeDispatched()` is a cold `flow {}` builder that collects a second `StateFlow` (`dispatched.first { it }`) before it reaches `emitAll(backing)`. Wrapped in `onStart` and `distinctUntilChanged`, that is several allocations plus an extra collector per subscription, all to reproduce `backing`.

```kotlin
combine(
    analytics.sourceConfigState.observeDispatched(),
    analytics.consentState.flow
) { sourceConfig, _ -> sourceConfig }
```

The source-config arm should stay on `observeDispatched()` — its seed is `SourceConfig.initialState()` and must be skipped.

Separately: the transform drops the consent value, and the resolver re-reads it live via `.value`. `combine` is a trigger here, not a data join. One comment saying so would save the next reader a detour.

vgupta98 (commented):
Reviewed this PR on its own diff against `feat/sdk-5259-consent-management-plugin`. The fail-open rules are clearly written and well covered by tests. The `filter` placed after the `combine` is the right call, and the comment explaining why is appreciated. Five notes inline.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/ConsentResolver.kt
SKannaniOS: Good catch — with no matching entry the resolver returns true and the destination is never gated, silently.

One correction though: custom isn't a limitation, it's the integration mechanism for any CMP. A OneTrust customer maps their decisions onto consent IDs and feeds them via setConsent, so they're supported. The narrower gap is a destination with consentManagement entries but no custom one.

Worth a diagnostic, but I'd add it with named-provider support (in V2 version) so the message can name the actual provider. Tracking it as a follow-up.

android/src/main/kotlin/com/rudderstack/sdk/kotlin/android/models/consent/ConsentResolver.kt
SKannaniOS: These are two different contracts — the plugin writes `context.consentManagement` onto the event, the resolver reads `config.consentManagement` from `SourceConfig`. Only provider overlaps; the rest of each block differs, so a wire change would only ever touch one of them.

Also #346 already drops the plugin's copy in favour of `SDKManagedContextKey.CONSENT_MANAGEMENT.key`. Leaving as is.

vgupta98 (approved):

vgupta98 (approved):

